PR #6, opened 2020-06-22
Stop using a build info file for build information

psoleckimoj:
- Remove build info code
- Pass build info into node app from env vars in Dockerfile
- Fix tests


server/config.js
```diff
@@ -52,6 +52,11 @@ module.exports = {
   test: isTest,
   production: isProduction,
   backendUrl,
+  buildInfo: {
```
LucasCairns, line 55: I'm wondering if we can be more explicit with these default values?

psoleckimoj: We need something fixed and immediately identifiable as not prod but otherwise open to suggestions 😄

LucasCairns: Other than `unset`, none immediately off the top of my head. But I agree we should focus on the fix at hand

PR #12, opened 2020-06-24
Allow popular topics section and title to be configurable per site

psoleckimoj:
- Move config into global config files
- Change title to be in config object


test/routes/index.spec.js
```diff
@@ -42,7 +42,7 @@ describe('GET /', () => {
     };
   });
 
-  describe('Homepage', () => {
+  describe.only('Homepage', () => {
```
LucasCairns, line 45: `.only` needs removing

psoleckimoj: Done! Test didn't work, now it does I shall remove.

PR #21, opened 2020-07-14
Update Helm to enable ingress

LucasCairns:
- Create helpers to generate host names
- Split site and environement config in values files
- Update Ingress and Notes template to use helper functions


helm_deploy/prisoner-content-hub-frontend/values.cw.yaml
```diff
@@ -0,0 +1,7 @@
+application:
+  config:
+    appName: "HMYOI Cookham Wood"
+    establishmentName: cookhamwood
+
+ingress:
+  domainPrefix: cw
```
spikeheap, line 7: Could we use `cookhamwood` (with or without hyphens) or something more descriptive in the name, domain, and references in `config.yml` here so the URL is meaningful?

.circleci/config.yml
```diff
@@ -264,8 +261,7 @@ jobs:
     steps:
       - release_to_namespace:
           namespace: "prod"
-          establishment: "cookhamwood"
-          establishmentDisplayName: "HMYOI Cookham Wood"
+          establishment: "cw"
           releaseName: "prisoner-content-hub-frontend-cw"
```
spikeheap, line 265: These two lines could do with updating to `cookhamwood` for consistency, and we should rename `values.cw.yaml` to `values.cookhamwood.yaml`

helm_deploy/prisoner-content-hub-frontend/values.cw.yaml
LucasCairns: Updated domain prefix

.circleci/config.yml
LucasCairns: Updated Circle and filename

PR #22, opened 2020-07-15
Add dev env Ingress

psoleckimoj:
- Enable the ingress in dev as documented in https://trello.com/c/Hm7FlZtE


psoleckimoj:
> Looks good 👍. Should we rename that values file to `development` rather than `non-prod`?

Sounds sensible, I'll do that but will need reapproving no doubt :D

PR #24, opened 2020-07-15
Feature/url rewrite feature toggle

LucasCairns:
Updated URL rewrites for media items to be behind a feature toggle, reasons for this are;

- Preserve default behaviours already present in Azure environments
- Allow Cloud Platform environment to be configured to use absolute URLS

### DIFF
--- a/server/utils/index.js
+++ b/server/utils/index.js
@@ -70,10 +70,7 @@ function relativeUrlFrom(url = '', override) {
 }
 
 function fixUrlForProduction(url, config = defaultConfig) {
-  if (config.useRelativeUrl) {
-    return relativeUrlFrom(url);
-  }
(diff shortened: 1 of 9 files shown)


server/utils/index.js
```diff
@@ -70,7 +70,7 @@ function relativeUrlFrom(url = '', override) {
 }
 
 function fixUrlForProduction(url, config = defaultConfig) {
-  if (config.production) {
+  if (config.useRelativeUrl) {
```
psoleckimoj, line 73: `return config.useRelativeUrl ? relativeUrlFrom(url) : url;` 😄

LucasCairns: Good shout, updated 👍

PR #35, opened 2020-07-23
Remove azure CI builds

spikeheap:
Following the call with Rich, here's a PR to remove the CI deploy to the Azure for all but Wayland and Berwyn production

There are some downsides to this PR – we won't have a non-production environment to recreate issues we see in Wayland and Berwyn.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,50 +103,7 @@ jobs:
       - run:
           name: Run Unit and Integration tests
           command: npm run lint && npm run test
-  preview_build:
-    <<: *defaults
-    steps:
-      - checkout
-      - setup_remote_docker:
-          docker_layer_caching: true
-      - restore_cache:
-          key: dependency-cache-{{ checksum "package-lock.json" }}
-      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -256,7 +202,6 @@ jobs:
             export VERSION_TO_DEPLOY="$(cat /tmp/build-info/version-to-deploy.txt)" && \
             ssh -t ciuser@pfs-management-bastion-1.uksouth.cloudapp.azure.com ssh -t ciuser@hub-wayland-prod.pfs-management.com "sudo /etc/docker-decomposed-secrets.sh -v ${VERSION_TO_DEPLOY}" && \
```
psoleckimoj, line 203: Don't these want to go to? We're not in Azure in prod either? Or do you just want to keep prod up-to-date for now?

spikeheap: Good point, this PR is a little out of date and we can get rid of even more now :+1:

spikeheap:
This isn't needed anymore :)

PR #39, opened 2020-07-30
Fix PDF downloads

LucasCairns:
PDF items were consistantly using relative URLs, this would break when we required absolute URLs, such as using pre-signed S3 links

- Update PDF route to use fixUrlProduction rather than relativeUrlFrom to allow behaviour to be consitant with other media
- Update PDF route to redirect rather than directly stream, as we can go direct to the PDF resource


spikeheap:
Just noticed the big play button doesn't have a hover style change, so it's not obvious it's clickable. As we're removing the control bar until they've hit play, it would be great to add an affordance for it :)

PR #40, opened 2020-08-04
Add Sentry exception logging, to centralise and alert on Node exceptions

spikeheap:
### Changes

Cloud Platform recently set up a Sentry organisation on a paid plan. [Sentry](https://sentry.io/for/node/) allows us to collect, track and notify us about code-level exceptions and events with stack-traces. 

This PR adds Sentry exception collection to the Node frontend. 

The `FRONTEND_SENTRY_DSN` has been added to CircleCI in development/staging/production, so exceptions should start appearing in the [dashboard](https://sentry.io/organizations/ministryofjustice/issues/?project=5377056).

We can test this by triggering an exception in a Kubernetes pod or locally (or just deliberately shipping unstable code 🤪). We should definitely validate that this happens as we expect in Cloud Platform.

### Intent

This will help us catch exceptions and errors early on, and goes hand-in-hand with the alerting & monitoring work we've got going on.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,5 +1,14 @@
 require('dotenv').config();
 
+const Sentry = require('@sentry/node');
+const { getEnv } = require('./utils/index');
+
+// Set up Sentry before (almost) everything else, so we can
+// capture any exceptions during startup
+Sentry.init({
+  dsn: getEnv('SENTRY_DSN', ''),
+});
+
 const app = require('./server/index');
 const { logger } = require('./server/utils/logger');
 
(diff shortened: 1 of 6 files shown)


server.js
```diff
@@ -1,5 +1,14 @@
 require('dotenv').config();
 
+const Sentry = require('@sentry/node');
+const { getEnv } = require('./utils/index');
+
+// Set up Sentry before (almost) everything else, so we can
+// capture any exceptions during startup
+Sentry.init({
+  dsn: getEnv('SENTRY_DSN', ''),
```
spikeheap, line 9: I added the `''` default so this won't break local development environments, but it'll raise an exception in production if it's not set.

PR #44, opened 2020-08-05
Limit video height to 500px

psoleckimoj:


server/views/pages/video.html
```diff
@@ -60,6 +60,7 @@ <h1 id="title" class="govuk-hub-media-player__title govuk-heading-xl govuk-body
         controls
         preload="auto"
         poster="{{ data.image.url }}"
+        style="max-height: 500px;"
```
spikeheap, line 63: I wonder if this could live as a CSS class rather than inline? 

Would it be worth us using `vh` to set this as a percentage of the height of the viewport to make this more flexible than a pixel count?

spikeheap (approved):
Minor comments around potential refactors, but nothing blocking this 👍

server/views/pages/video.html
psoleckimoj: @spikeheap I should've started as a Draft anyways 🤦 It doesn't work anyway.

PR #49, opened 2020-08-11
Add missing approval step in Circle config for deploy Staging

LucasCairns:


spikeheap:
I think we should automatically deploy the latest `main` to the staging environment. This feels like a regression on the delivery part of CI/CD. I'll pick up the conversation in Slack :)

PR #50, opened 2020-08-11
Update Circle pipeline

LucasCairns:
- Removed deploy step to Development for the main branch
- Re-implemented unit and integration tests for all branches


spikeheap:
And we saw it work in Circle on this branch

PR #54, opened 2020-08-12
Parameterise CircleCI jobs

spikeheap:
This is a bit of an exploration. The PR:

- Refactors code checkout to occur in the job step
- Uses relative paths to remove the need for a custom checkout location
- [more controversially] uses `matrix` parameterisation to deduplicate the release steps. 

The result (after a bit of jigerry-pokery) is the following, where the job name becomes more meaningful in the UI:

Screenshot 2020-08-12 at 15 48 22

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,57 +172,48 @@ jobs:
 
   deploy_cloud_platform_production:
     <<: *defaults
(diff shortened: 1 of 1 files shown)


psoleckimoj (approved):
LGTM!

PR #64, opened 2020-08-26
Rename new designs feature flag

psoleckimoj:
- FEATURE_NEW_DESIGNS becomes ENABLE_PERSONAL_INFORMATION
- newDesigns in the config object becomes personalInformation

### DIFF
--- a/server/views/pages/home.html
+++ b/server/views/pages/home.html
@@ -56,7 +56,7 @@
         </ul>
       </div>
     </div>
-    {% if config.newDesigns %}
+    {% if config.personalInformation %}
       {{personalScheduleToday({events: todaysEvents, isTomorrow: isTomorrow, establishmentId: config.establishmentId}, config.userName)}}
     {% endif %}
   </div>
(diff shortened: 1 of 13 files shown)


server/views/pages/home.html
```diff
@@ -56,7 +56,7 @@ <h3 class="govuk-heading-s">{{ homePageLinksTitle }}</h3>
         </ul>
       </div>
     </div>
-    {% if config.newDesigns %}
+    {% if config.personalInformation %}
```
spikeheap, line 59: Really trivial thought: should we carry through the name, to make it more obvious it's a toggle/boolean rather than an object holding personal information?

Handily, naming that would be tricky. `config.enablePersonalInformation` almost looks like it could be a function to enable personal information. `config.PersonalInformationEnabled` might fit better but doesn't match the environment variable name 🤷.

Definitely not blocking for this PR, just an early morning musing 🙂.

spikeheap (approved):
This looks good. I like that we've gone with `ENABLE_PERSONAL_INFORMATION`. You're absolutely right that it's neater to keep the positive assertion of `ENABLE_` consistent for feature toggles. The other way is a bit janky.

server/views/pages/home.html
psoleckimoj: Completely understand, the latter would be best, maybe a tech debt thing to look at after 😄

PR #80, opened 2020-09-02
Use single GAID per env rather than per site

psoleckimoj:
- Change Helm config to support new properties
- Change default in config


LucasCairns (approved):
LGTM :shipit:

PR #83, opened 2020-09-02
update package lock

davidthomas22:
*Update minor dependencies
*Update bug in returns a topics menu test


psoleckimoj:
Hmm, this is returning different results in Circle to local again...

psoleckimoj (approved):
LGTM!

PR #84, opened 2020-09-03
Return textual names for secondary tags and categories

psoleckimoj:
- Parse existing secondary tags into a string
- Add categories in the same way as secondary tags
- Amend data sent to GA
- Amend test


LucasCairns (approved):
👍

LucasCairns (approved):
:shipit:

PR #85, opened 2020-09-03
Refactor offenderService and increase test coverage

LucasCairns:


server/routes/index.js
```diff
@@ -36,17 +43,32 @@ const createIndexRouter = ({ logger, hubFeaturedContentService, config }) => {
...
         personalInformation,
-        userName: userName ? `Hi, ${userName}` : null,
         establishmentId,
         returnUrl: req.originalUrl,
       };
 
+      const data = {};
+
+      if (req.user) {
+        const userName = req.user && `Hi, ${req.user.getFullName()}`;
```
psoleckimoj, line 53: `Hi, ` isn't part of their username, it's a display thing so should be separate. Either in the template or a separate variable.

server/utils/string.js
```diff
@@ -0,0 +1,20 @@
...
+  return amount && currency
+    ? new Intl.NumberFormat('en-GB', {
+        style: 'currency',
+        currency,
+      }).format(amount)
+    : placeHolder;
+};
+
+module.exports = {
+  fullNameOr,
```
psoleckimoj, line 18: Not gonna not LGTM this but I've never been a fan of function names ending half way through a sentence as they aren't self descriptive, as in Or what? They only make sense if you know what the args are. Same with For something as well, you're passing something in so it has to be For those args anyway.

server/routes/index.js
psoleckimoj: Actually, this would be better if we rename the var being passed in to `userNameDisplay` or something to indicate that?

LucasCairns: Resolved, updated to `welcomeMessage`

server/utils/string.js
LucasCairns: Renamed to `..OrDefault` to be more explicit with intent

psoleckimoj (approved):
LGTM!

PR #86, opened 2020-09-07
Remove jplayer from tag template

psoleckimoj:
- Fix test
- Remove jplayer code
- Remove component used in only this template and move remaining code to template


LucasCairns (approved):
LGTM :shipit:

PR #87, opened 2020-09-07
Replace jplayer with videojs in flat content

psoleckimoj:
- Remove jplayer code and module
- Add videojs to flat content
- Replace jplayer html and js with equivalent videojs code


LucasCairns (approved):
👍

PR #89, opened 2020-09-08
Remove jplayer from flat content

psoleckimoj:
- Remove video matching
- Remove video display in template
- Uninstall jplayer
- Remove jplayer skin


LucasCairns (approved):
:shipit:

PR #94, opened 2020-09-10
Remove unused selectors

psoleckimoj:
- Most are no longer used
- Found used ones by finding the file include which shows only 3 are used


davidthomas22 (approved):
LGTM

PR #112, opened 2020-09-15
Switch config based on host

psoleckimoj:
Initial test

- Switches most config in app.js
- Need to get switched config into server/clients/hub.js


psoleckimoj:
Tested this using local host file entries:

http://wayland-prisoner-content-hub-local:3000/
http://berwyn-prisoner-content-hub-local:3000/

server/app.js
```diff
@@ -176,13 +175,20 @@ const createApp = ({
...
 
-  // GovUK Template Configuration
-  const establishmentId = getEstablishmentId(config.establishmentName);
-  app.locals.asset_path = '/public/';
-  app.locals.config = {
-    ...config,
-    establishmentId,
-  };
+  app.use((req, res, next) => {
+    const replaceUrl = /-prisoner-content-hub.*$/g;
```
psoleckimoj, line 179: This string could be put into config I guess as you can do `new RegExp` from a string.

spikeheap: This looks great, and would work nicely for review apps as well if we sit them under a domain for each PR/deployment 🙌

PR #116, opened 2020-09-17
Make failed auth url configurable

psoleckimoj:
- Add config to Helm
- Add config to config file
- All url to be injected to allow testing

### DIFF
--- a/helm_deploy/prisoner-content-hub-frontend/templates/deployment.yaml
+++ b/helm_deploy/prisoner-content-hub-frontend/templates/deployment.yaml
@@ -110,6 +110,8 @@ spec:
                   key: azureAdClientSecret
             - name: AZURE_AD_CALLBACK_URL
               value: {{ printf "%s%s" (include "prisoner-content-hub-frontend.externalHost" .) .Values.prisonerAuthAD.callbackEndpoint }}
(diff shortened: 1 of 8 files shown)


spikeheap:
Do we want to tightly couple our auth flow with the CMS? I can see it has some benefits by allowing other members of the team to add content, but it _feels_ like a frontend system flow, and feels distinct from CMS content. 

As an example, I would expect the auth failure to be reported correctly if we switched to a new CMS, or had a test environment with no content.

psoleckimoj:
@spikeheap it's configurable? It's not tied to anything 😕

davidthomas22 (approved):
LGTM

spikeheap:
> @spikeheap it's configurable? It's not tied to anything 😕

What I'm getting at is that it won't work without those CMS pages, so our frontend auth flow demands a specific piece of content exist in the CMS for the auth flow to work (or rather, fail) correctly.

What makes me uncomfortable is that we're taking something that (I believe) is an integral part of the frontend auth flow, and shifting that concern into the CMS. 

As an example, we've talked about being able to fire up a vanilla version of the Drupal instance for development/test environments. To have a functional (albeit empty) site, this change introduces the need for us to create at least one page.

PR #123, opened 2020-09-21
Base establishment on hostname

psoleckimoj:
- Switches most config in app.js
- Need to get switched config into server/clients/hub.js

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -162,7 +162,7 @@ const createApp = ({
 
   app.use((req, res, next) => {
     const replaceUrl = /-prisoner-content-hub.*$/g;
-    config.establishmentName = pathOr('berwyn', ['headers', 'host'], req)
+    config.establishmentName = pathOr('wayland', ['headers', 'host'], req)
       .split('.')[0]
       .replace(replaceUrl, '');
     const establishmentId = getEstablishmentId(config.establishmentName);
--- a/server/app.js
+++ b/server/app.js
@@ -161,18 +161,28 @@ const createApp = ({
   );
 
   app.use((req, res, next) => {
-    const replaceUrl = /-prisoner-content-hub.*$/g;
-    config.establishmentName = pathOr('wayland', ['headers', 'host'], req)
-      .split('.')[0]
-      .replace(replaceUrl, '');
-    const establishmentId = getEstablishmentId(config.establishmentName);
+    if (req.session && !req.session.id) {
+      const replaceUrl = /-prisoner-content-hub.*$/g;
+
+      req.session.id = uuid();
+      req.session.establishmentName = pathOr(
+        'wayland',
+        ['headers', 'host'],
+        req,
+      )
+        .split('.')[0]
+        .replace(replaceUrl, '');
+    }
+
+    res.locals.feedbackId = uuid();
+
     app.locals.asset_path = '/public/';
     app.locals.config = {
       ...config,
-      establishmentId,
+      establishmentId: getEstablishmentId(req.session.establishmentName),
     };
 
-    return next();
+    next();
   });
 
   // Don't cache dynamic resources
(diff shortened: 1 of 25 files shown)


psoleckimoj:
Checked on dev with backend fix and works.

server/app.js
```diff
@@ -176,13 +175,20 @@ const createApp = ({
...
 
-  // GovUK Template Configuration
-  const establishmentId = getEstablishmentId(config.establishmentName);
-  app.locals.asset_path = '/public/';
-  app.locals.config = {
-    ...config,
-    establishmentId,
-  };
+  app.use((req, res, next) => {
+    const replaceUrl = /-prisoner-content-hub.*$/g;
```
spikeheap, line 179: I'm loving where this is going 🙌. Would it be worth us adding a line to the readme that calls out that we do hostname splitting for staging/dev environments so the expectation for deploys is obvious?

psoleckimoj: Makes sense 😄

psoleckimoj: Done 😄

server/app.js
```diff
@@ -165,13 +164,20 @@ const createApp = ({
...
+    config.establishmentName = pathOr('wayland', ['headers', 'host'], req)
```
LucasCairns, line 169: Would there be an issue with mutating the config? could we use a local variable?

LucasCairns: This would become a problem when merge the frontend instances in to a single deployment, mutating the config would affect other requests mid flight

psoleckimoj: Then it needs a complete rethink :( Maybe establishmentName needs to be added into the req object? The requests need to know about the establishment name.

psoleckimoj: This only ever seems to be used when setting the session so maybe we need to look at the ordering and check the session id? If it's the same session then we don't attempt to change the config? Or something?

psoleckimoj: Or we put it in the session? Might make more sense? If it's not present then it must be a new session so we set the establishment.

LucasCairns: We could put it in the session, to avoid regenerating all the time. However it may cause issues, for when we switch between environments

LucasCairns: I think putting both `establishmentId` and `establishmentName` on the `req` object would work though

psoleckimoj: They're already in the req object as part of config though :( Sessions shouldn't carry across environments as they're different sites?

davidthomas22 (approved):
LGTM

LucasCairns (approved):
LGTM

server/app.js
```diff
@@ -161,14 +158,6 @@ const createApp = ({
     ),
   );
 
-  // GovUK Template Configuration
-  const establishmentId = getEstablishmentId(config.establishmentName);
-  app.locals.asset_path = '/public/';
```
psoleckimoj, line 166: This doesn't seem to be used from a search across the codebase.

server/services/hubContent.js
```diff
@@ -10,13 +10,7 @@ function createHubContentService({
...
     }
 
-    const content = await contentRepository.contentFor(id);
-    const prisonIds = prop('establishmentIds', content);
-
-    if (!canAccessContent(establishmentId, prisonIds)) {
-      return {};
-    }
-
+    const content = await contentRepository.contentFor(id, establishmentId);
```
psoleckimoj, line 13: Access is now controlled in the backend app as it should be.

server/repositories/hubContent.js
```diff
@@ -14,15 +14,18 @@ const {
...
     if (!id) {
       logger.error(`HubContentRepository (contentFor) - No ID passed`);
       return null;
     }
+    const query = {
+      _prison: establishmentId,
+    };
 
-    const response = await httpClient.get(endpoint);
+    const response = await httpClient.get(endpoint, { query });
```
psoleckimoj, line 28: Endpoint was changed to use the prison, previously it was ignored.

LucasCairns (approved):
LGTM 👍

PR #127, opened 2020-09-22
Remove font sizing from error message

LucasCairns:
- Remove font sizing from error message to be consistant with 404 page


psoleckimoj (approved):
LGTM

PR #143, opened 2020-09-29
Refactor design of balance component

LucasCairns:
balances

### DIFF
--- a/assets/javascript/all.js
+++ b/assets/javascript/all.js
@@ -14,14 +14,14 @@ function sendPageTrack(data) {
   });
 }
 
-(function() {
-  document.body.addEventListener('click', function(event) {
+(function () {
+  document.body.addEventListener('click', function (event) {
     if (matchesAll('[data-state]', event.target)) {
       var element = getAncestorElementBySelector('[data-state]', event.target);
 
       if (element) {
         var state = element.getAttribute('data-state');
-        var stateToggle = function() {
+        var stateToggle = function () {
           element.setAttribute('data-state', 'hidden');
         };
         element.setAttribute('data-state', toggleShowing(state));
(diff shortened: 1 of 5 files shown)


assets/javascript/balances.js
```diff
@@ -0,0 +1,25 @@
...
+    const HIDE_LABEL = 'Hide balances';
+    const PLACEHOLDER = '****';
+    balanceToggle.innerHTML = SHOW_LABEL;
+    const balances = document.getElementsByClassName('balance-amount');
+    for (let i = 0; i < balances.length; i++) {
+      balances[i].innerHTML = PLACEHOLDER;
+      balances[i].setAttribute('aria-hidden', true);
+    }
+    balanceToggle.onclick = e => {
+      const isHidden = e.target.innerHTML === HIDE_LABEL;
```
psoleckimoj, line 15: Would it not be better to check the `aria-hidden` attribute? Feels better than checking the HTML.

LucasCairns: The `aria-hidden` is on the balances sections, not the button (`e.target`). Maybe we could store a local variable for state?

psoleckimoj: Doh, sorry, nah leave as is in that case.

spikeheap:
Here's a card to track this :) https://trello.com/c/rxkGwTQl/1746-refactor-design-of-balance-component-historical

assets/javascript/all.js
```diff
@@ -84,7 +84,7 @@ function showHiddenBlock(block) {
     userAgent: navigator.userAgent,
   });
 
-  window.setTimeout(function() {
+  window.setTimeout(function () {
     button.style.display = 'block';
     hiddenData.style.display = 'none';
   }, 3000);
```
spikeheap, line 90: Looking at https://dsdmoj.atlassian.net/wiki/spaces/HUB/pages/2475295001/Viewing+account+balances+on+The+Content+Hub if we choose/chose option 1 I think we need to remove this timeout

LucasCairns: This code is the old code which used originally for balances, this is still used on the `Visits` page for displaying the `Visitor Name` so has not been removed. I think the only reason this file has come up in this PR is because of the formatting changes

psoleckimoj (approved):
LGTM

PR #147, opened 2020-10-01
Remove Getting A Job

psoleckimoj:


davidthomas22 (commented):
LGTM

davidthomas22 (dismissed):
LGTM

server/config.js
```diff
@@ -34,7 +31,6 @@ const establishments = {
     prefix: 'HMP',
     uuId: 'fd1e1db7-d0be-424a-a3a6-3b0f49e33293',
     standFirst: 'What you need to do to get or change your job in Berwyn.',
```
LucasCairns, line 33: is `.standFirst` related to GAJ? could we remove?

psoleckimoj: Good spot!!! Will remove :)

psoleckimoj: Actually, it appears to be used in the flat_content template and for that it doesn't use a path from config so I'll just remove the getting a job bits :)

LucasCairns: Awesome 💯

psoleckimoj: It's also in the basic page template.

psoleckimoj: Done

PR #148, opened 2020-10-01
Create codeql-analysis.yml

psoleckimoj:
Following @andrewrlee example, I thought it'd be interesting to see what happens 😄

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -47,7 +47,7 @@ jobs:
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
-        # By default, queries listed here will override any specified in a config file. 
+        # By default, queries listed here will override any specified in a config file.
         # Prefix the list here with "+" to use these queries and those in the config file.
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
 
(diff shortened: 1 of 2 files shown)


spikeheap:
👍 this is great. If we're taking this on, it'd be great to update our Confluence with a lightweight developer guide with links like this. 

We'll also need to decide which services we're going to live by – there's the Sonarqube instance we might want to consider, and the results from CodeQL look pretty actionable. It's also really nice that CodeQL can be a PR check, if it doesn't have too many false-negatives that could be very handy.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [main]
+  schedule:
+    - cron: '0 4 * * 0'
```
spikeheap, line 15: I guess a schedule is useful when we don't have changes but the CodeQL rules are updated?

spikeheap (approved):

.github/workflows/codeql-analysis.yml
psoleckimoj: @spikeheap I just copied the default code like it said 😄 We can make any changes we need to.

PR #153, opened 2020-10-07
Add real front-end GA code for testing

psoleckimoj:
- Real code required for checking from a webpage
- Only page tracks and using dev account so as not to ruin current analytics


spikeheap (approved):
:+1: ❤️ that that analytics key is our development estate

PR #154, opened 2020-10-08
Fix page titles

psoleckimoj:
- Homepage should be Home
- Tags should be the title amnd not establishment name


LucasCairns (approved):
LGTM

davidthomas22 (commented):
LGTM

PR #156, opened 2020-10-08
Apply a maximum length to search box

psoleckimoj:


LucasCairns (commented):
If we're doing this could we also include server-side validation?

psoleckimoj:
@LucasCairns we could indeed, I'll leave it for when we get onto the card 😄

davidthomas22 (approved):
LGTM

PR #158, opened 2020-10-09
Feature/increase logging in personalisation

LucasCairns:


server/routes/incentives.js
```diff
@@ -20,14 +20,14 @@ const createIncentivesRouter = ({ hubContentService, offenderService }) => {
         returnUrl: req.originalUrl,
       };
 
-      if (req.user) {
-        const userName = req.user && req.user.getFullName();
-        const { bookingId } = req.user;
+      const { user } = req;
+
+      if (user) {
```
psoleckimoj, line 25: Is this required? Won't the line above fail if `user` doesn't exist?

LucasCairns: Shouldn't.. AFAIK it would fail if req was undefined, if a user doesn't exist it would be undefined on `ln:25` so would pass over the `if` block

psoleckimoj: 👍

psoleckimoj (approved):
LGTM

PR #161, opened 2020-10-22
Move site config to single frontend

psoleckimoj:
- Remove establishment config files
- Change frontend to be site agnostic
- Move sitenames into values and loop round

NOTE: This requires the old `prisoner-content-hub-*` uninstalling with Helm.


helm_deploy/prisoner-content-hub-frontend/values.yaml
```diff
@@ -1,8 +1,8 @@
 nameOverride: ""
 fullnameOverride: ""
 
-replicaCount: 2
-maxReplicaCount: 3
+replicaCount: 3
+maxReplicaCount: 6
```
psoleckimoj, line 5: Up for discussion on these....

spikeheap: I'm keen to have this driven by the results of the performance test. Part of me wonders if we'd actually cope with fairly high load with a small number of pods seeing as the node process is fairly lightweight & fast, but we can definitely make better decisions with some data. 

We should be able to say "adding 1 pod to the frontend increases capacity by `m` users" (approx, and with all the caveats) and judge how & when to add them. It feels like we need a "brave new world" gif 🤪.

helm_deploy/prisoner-content-hub-frontend/templates/ingress.yaml
```diff
@@ -15,18 +15,24 @@ spec:
...
   {{- range .Values.ingress.hosts }}
-    - host: {{ $domainPrefix }}{{ .suffix }}
+    {{- $domainSuffix := .suffix -}}
+    {{- range $.Values.ingress.prefix }}
+    - host: {{ .siteName }}{{ .$domainSuffix }}
       http:
         paths:
           - path: "/"
             backend:
               serviceName: {{ $fullName }}
```
psoleckimoj, line 34: Not sure if this is OK to stay the same?

helm_deploy/prisoner-content-hub-frontend/values.yaml
psoleckimoj: Yeah it's a difficult one , we could set to 6 as that's what we've got now but those 6 will be shared between all 3 sites.

helm_deploy/prisoner-content-hub-frontend/templates/ingress.yaml
psoleckimoj: Seems to be fine...

helm_deploy/prisoner-content-hub-frontend/values.yaml
psoleckimoj: See Lucas' notes on what we managed with testing. 3 will be fine for now.

helm_deploy/prisoner-content-hub-frontend/templates/_helpers.tpl
```diff
@@ -73,7 +73,7 @@ Create external Kubernetes hostname
 {{- if .Values.ingress.tlsEnabled }}
 {{- $protocol = "https" }}
 {{- end }}
-{{- printf "%s://%s%s" $protocol .Values.ingress.domainPrefix (index .Values.ingress.hosts 0).suffix }}
+{{- printf "%s://siteName%s" $protocol (index .Values.ingress.hosts 0).suffix }}
```
psoleckimoj, line 76: Up for discussion....

davidthomas22 (dismissed):
LGTM

helm_deploy/prisoner-content-hub-frontend/templates/_helpers.tpl
spikeheap: I like the idea of renaming `domainPrefix` to `siteName`, that really makes the intent of that variable clearer 🙌 .

Looking at the code change, I'm probably reading it wrong (many hours into an assessment call), but I think that'll render `siteName` as a string, rather than interpolating the value.

spikeheap: I wonder if we could print out the entire list of site URLs, that'd be a pretty nifty validation that of the live sites at the end of the deployment?

psoleckimoj: `siteName` is replaced when creating an strategy, we can't do it at this stage as it's not known until a request is received.

psoleckimoj: Like the last suggestion though 😄

server/app.js
```diff
@@ -120,21 +120,30 @@ const createApp = ({
...
+  app.use((req, res, next) => {
+    passport.use(
+      new AzureAdOAuth2Strategy(
+        {
+          clientID: config.auth.clientId,
+          clientSecret: config.auth.clientSecret,
+          callbackURL: config.auth.callbackUrl.replace(
+            'siteName',
+            req.session.establishmentName,
+          ),
```
spikeheap, line 139: Are we able to get the hostname from the request, perhaps from a header passed by the load balancer? It _feels_ like doing a find/replace on the callbackUrl might catch us out later on because it's not what you'd expect when you first approach it.

It looks like we need the `callbackPath` as an environment variable, and append that to the hostname we derive automatically.

helm_deploy/prisoner-content-hub-frontend/templates/_helpers.tpl
spikeheap: It feels a bit odd to be printing a URL that doesn't work here. I think I got tripped up because `siteName` isn't obviously a variable to be used elsewhere. 

I've added a comment on the point that does the replacement, and it feels like we can make that more elegant/explicit by using a different environment variable. I wonder if we'd use this output to set the `callbackUrl`? If not, we could just print the list of all hostnames, and add documentation around CI/helm, or address that other point?

server/app.js
LucasCairns: We do that when we infer the establishment in `configureEstablishment.js - line 18`
```
    const establishmentName = pathOr('wayland', ['headers', 'host'], req)
      .split('.')[0]
      .replace(replaceUrl, '');
```
So maybe we could `req.headers.host` here?

psoleckimoj: @LucasCairns we do? it just has a `pathOr` on it. `host` includes the full hostname not just `berwyn` etc.

psoleckimoj: Ignore me, read the rest of your comments, hopefully the PR is better and cleaner now.

LucasCairns (dismissed):
LGTM 👍

spikeheap (approved):
Looks great to me

LucasCairns (approved):
LGTM :shipit:

PR #167, opened 2020-10-28
Enable overall personalisation in prod

psoleckimoj:
- Per site is now in config.js


davidthomas22 (approved):
LGTM

LucasCairns (approved):

spikeheap:
I think we can actually just remove the deployment-level feature toggle. We're able to toggle at a prison level, so I _think_ this is complexity we don't need in the deployment?

PR #168, opened 2020-10-29
Fix ingress layout

psoleckimoj:


helm_deploy/prisoner-content-hub-frontend/templates/ingress.yaml
```diff
@@ -14,11 +14,11 @@ metadata:
 spec:
   tls:
   {{- range .Values.ingress.hosts }}
+  {{- $certSecret := .cert_secret -}}
```
psoleckimoj, line 17: I understand more formatting now and I always wanted these here 😁

spikeheap: Ahhh yeah, I've been bitten by that a couple of times 🤦. Elegant helm indentation != valid YAML ☹️.

davidthomas22 (approved):
LGTM

spikeheap (approved):
Looks good, and I've felt that pain! 

I _think_ the indentation on L21 was correct (it's 2 deeper than the previous item), but YAML copes with the extra level as long as it's consistent so whichever feels nicer is good ✅

PR #177, opened 2020-11-18
Update Andrew at NPRs IP

psoleckimoj:


spikeheap (approved):
:+1: looks reasonable. Is there a card for this? I wonder if we could link to something that would let us validate the IP for typos in any future ones? Not blocking for this though :)

PR #179, opened 2020-11-25
Send prisonId to vocabulary/tag API

psoleckimoj:


LucasCairns (commented):
Not sure if it's already been done, but have we updated the tests to reflect this change?

psoleckimoj:
> Not sure if it's already been done, but have we updated the tests to reflect this change?

There are no tests for this endpoint 🤷‍♂️ I could add some I guess.

psoleckimoj:
> > Not sure if it's already been done, but have we updated the tests to reflect this change?
> 
> There are no tests for this endpoint 🤷‍♂️ I could add some I guess.

Tests added for current functionality, once the endpoint itself changes we can add tests, otherwise it's not testing current functionality.

LucasCairns (approved):
LGTM 👍

server/repositories/__tests__/hubMenuRepository.spec.js
```diff
@@ -67,4 +68,34 @@ describe('hubMenuRepository', () => {
...
+        },
+      ];
+
+      const result = await repository.tagsMenu({
+        prisonId: 2,
+      });
+
+      expect(result).toStrictEqual(expected);
+    });
+  });
```
spikeheap, line 100: I'm probably missing something, but does this test assert that the request to the API includes the `prisonId`?

psoleckimoj: I suppose we could assert on the mock that it's been called and with the `prisonId`?

We're not actually calling the API here, we're calling `httpClient.get()` which builds the query and sends it off.

psoleckimoj: I could amend all the tests in this since none of them do that either. Or it can get looked at as tech debt.

spikeheap: That's an interesting one. You're right, asserting the mocked `httpClient` call includes the `prisonId` would give us a test which would have failed before the change in this PR 👍.

As for all the tests, I guess if we have at least one including the `prisonId` we've got coverage. It's not clear cut, but I tend to prefer to test what mocks have been called with as that's the boundary we're introducing for the test... though sometimes that can be more effort than it's worth 🤷 .

PR #180, opened 2020-11-25
Fix missing query parameter when calling the Term endpoint

LucasCairns:
The fix is this

```
const response = await httpClient.get(endpoint, { query });
```

Originally we were passing the query as the second parameter, which is wrong, it should be wrapped in an object

So `{ _prison: 123 }` becomes `{ query: { _prison: 123 } }`


psoleckimoj (approved):
LGTM!

PR #188, opened 2020-12-10
Upgrade node modules

psoleckimoj:
- New linting rules introduced
- Fix linting problems


server/auth/middleware.js
```diff
@@ -97,7 +95,6 @@ const createSignOutMiddleware = ({ logger, analyticsService }) => {
     });
     res.redirect(req.query.returnUrl || '/');
```
LucasCairns, line 96: Regarding the security advisory here, we could put a check in that the URL is relative, or ensure it is relative. So either;
`if (url.indexOf('://') > 0 || url.indexOf('//') === 0)`
or
`Url.parse(url).path()`

psoleckimoj: No security issues on this PR 🤷‍♂️ 

it can get done separately.

psoleckimoj: Well, there's a warning

psoleckimoj: Has someone reenabled this or something as it's now picking up stuff that's been in forever as security issues? Which is good but bad that they've not been picked up before.

LucasCairns: Ah okay, it showed up for me as a warning. I'll add it as a card to the Tech Debt column 👍

psoleckimoj: @LucasCairns I'll card up this issue. It's not new and it's not failing the build. If you can check the PR for the change it's supposed to contain please.

LucasCairns (approved):
LGTM 👍

PR #190, opened 2020-12-16
Add Christmas crossword

LucasCairns:
### Context

https://trello.com/c/hADmC1bR/1760-create-christmas-crosswords-schedule-release

### Intent

Added the route `/games/christmas-crossword` and applied the initial list of Christmas clues

Screenshot 2020-12-16 at 11 54 49

### Considerations

This change simply copies existing code for the crossword and modifies the clue list to be Christmas theme, this means that we should avoid major refactoring and presumably retain the same amount of test coverage.

N/A

### Checklist

- [ ] Tested in Development

### DIFF
--- a/assets/javascript/games/crossword/christmas.js
+++ b/assets/javascript/games/crossword/christmas.js
@@ -12,9 +12,7 @@ var WORD_LIST = [
   { answer: 'holly', clue: 'Spiky leaves' },
   { answer: 'tinsel', clue: 'Shiny on the tree' },
   { answer: 'tree', clue: 'Real ones leave a mess on the floor' },
-  {
-    answer: 'twelfth', clue: 'It\'s considered unlucky to have decorations up after this time - _______ Night'
-  },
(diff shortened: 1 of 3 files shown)


psoleckimoj:
Maybe not the right time to do it but seeing as the only change is the word list and this is just on a different route, is there no way to pull that out into a config file and just change the config based on route?

LucasCairns:
Definitely something we should look in to, but for this PR I have tried to avoid any major code changes, so we can simply throw this away after the festive period

spikeheap (approved):
This looks good :+1:. I've tested it on dev and the success/reset flow works nicely.

:+1: to copy/pasting to reduce the effort right now. That said, it'd be worth throwing a card into Trello to cover that refactor so next year we can just add a route/question list if we need a new crossword. Great suggestion @psoleckimoj

psoleckimoj (dismissed):
LGTM!

davidthomas22 (approved):
LGTM

davidthomas22 (approved):
LGTM

PR #203, opened 2021-02-11
Add feature toggle for mocking auth

LucasCairns:
### Context

https://trello.com/c/7XkW9uMD/1931-how-do-we-test-sign-in-out-e2e

### Intent

- Create mock auth middleware
- Add feature toggle `ENABLE_MOCK_AUTH`

So.....

- Clicking sign-in/out with mocking enabled will use a "test" user

### Considerations

Turns out mocking getting our AzureAD SSO set up with Cypress is difficult, especially with our current setup and how we intend to run these tests.
To ensure we're not blocked on this work, and until we have a better solution, there's this PR ☝️ 

### Checklist

- [x] Tested in Development

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -138,6 +138,7 @@ module.exports = {
     personalInformation:
       getEnv('ENABLE_PERSONAL_INFORMATION', 'false') === 'true',
     useRedisCache: getEnv('ENABLE_REDIS_CACHE', 'true') === 'true',
+    useMockAuth: getEnv('ENABLE_MOCK_AUTH', 'false') === 'true',
     showStackTraces:
       getEnv('ENABLE_STACK_TRACES_ON_ERROR_PAGES', 'false') === 'true',
   },
--- a/helm_deploy/prisoner-content-hub-frontend/templates/deployment.yaml
+++ b/helm_deploy/prisoner-content-hub-frontend/templates/deployment.yaml
@@ -92,6 +92,8 @@ spec:
               value: {{ .Values.application.sentry_dsn }}
             - name: ENABLE_REDIS_CACHE
               value: {{ .Values.application.config.enableRedisCache | quote }}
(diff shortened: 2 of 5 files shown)


server/config.js
```diff
@@ -138,6 +138,7 @@ module.exports = {
     personalInformation:
       getEnv('ENABLE_PERSONAL_INFORMATION', 'false') === 'true',
     useRedisCache: getEnv('ENABLE_REDIS_CACHE', 'true') === 'true',
+    useMockAuth: getEnv('ENABLE_MOCK_AUTH', 'false') === 'true',
```
spikeheap, line 141: Would it be worthwhile adding an explicit check for production here, so it refuses to mock auth in prod?

spikeheap (approved):
:+1: Do you think there are any tests we might want to add around this to assert that auth is mocked when we want it to be?

server/config.js
LucasCairns: Yeeah, will sort out some tests and see what I can do about doing that additional check 👍 Just wanted to get something out so we're not blocking writing E2E's for personalisation features

davidthomas22 (approved):
LGTM

server/config.js
LucasCairns: I've added some tests to ensure the correct middleware gets mounted based on the feature toggle being set

spikeheap: Hi @LucasCairns, did we update this to explicitly prevent mocked auth in production? It looks like the response to this comment might have been addressing a different point.

LucasCairns: I've set `ENABLE_MOCK_AUTH` to be explicitly `false` in our Helm templates. We don't have variable that says what environment we're in, but we have the `values` files for each environment. So we can split this out down the line if we choose to have mock auth in `Development` for example.

PR #205, opened 2021-02-16
Add transactions page

davidthomas22:
### Context

https://trello.com/c/2XLGUCQB

### Intent

-Add transactions page
-Update cypress tests

### Checklist

- [x] Tested in Development


swilson-mojhub:
This is interacting with an API? should we have integration tests?

LucasCairns:
I've covered the the call to the API and mapping/handling of responses with units. I'll add some tests to cover the integrations between the repository (API call) and the route, I'm assuming that's what you mean?

server/services/offender/responses/transaction.js
```diff
@@ -0,0 +1,56 @@
...
+      currentBalance,
+      entryDescription,
+      agencyId,
+      currency,
+    } = options;
+    this.paymentDate = entryDate;
+    this.postingType = postingType;
+    this.amount = penceAmount / 100;
+    this.currency = currency;
+    this.balance = currentBalance / 100;
```
spikeheap, line 23: Could we store currency amounts as pence until the last possible moment, and do decimal formatting in the presentation layer? Casting early runs the risk of dealing with rounding errors, so we could trip ourselves up by casting early if we used these values for any arithmetic in future.

server/services/offender/index.js
```diff
@@ -247,6 +277,27 @@ const createOffenderService = (
...
+      const response = await repository.getPrisonDetailsFor(prisonId);
+      return Prison.from(response).format();
+    } catch (e) {
+      logger.error(
+        `OffenderService (getPrisonDetailsFor) - Failed: ${e.message} - prisonId: ${prisonId}`,
+      );
+      logger.debug(e.stack);
+      return null;
+    }
+  }
```
spikeheap, line 299: Mirroring the comment on `getTransactionsFor` (but GitHub won't let me link because it's a draft review)... What motivates the split between a repository and service here? 

Could we keep the response error-handling logic alongside the API call in the repository? Splitting the two means we're bleeding concerns out into surrounding functions and tightly coupling them.

server/routes/money.js
```diff
@@ -38,6 +39,84 @@ const createMoneyRouter = ({ hubContentService, offenderService }) => {
...
+          (accumulator, prison) => {
+            if (Object.hasOwnProperty.call(accumulator, prison.prisonId)) {
+              return accumulator;
+            }
+            const updated = { ...accumulator };
+            updated[prison.prisonId] = prison.longDescription;
+            return updated;
+          },
+          {},
+        );
```
spikeheap, line 98: Is `reduce` the most appropriate approach here? It feels like there's a lot of complexity in those lines, and I wonder if there's a more straightforward way of achieving the same outcome.

server/routes/money.js
```diff
@@ -38,6 +39,84 @@ const createMoneyRouter = ({ hubContentService, offenderService }) => {
...
+            accountTypeMachineName[accountType],
+            formatISO(transactionsFrom, { representation: 'date' }),
+            formatISO(transactionsTo, { representation: 'date' }),
+          ),
+          offenderService.getBalancesFor(user),
+        ]);
+
+        const listOfPrisons = Array.from(
+          new Set(transactions.map(transaction => transaction.prison)),
+        );
```
spikeheap, line 82: This looks nifty. What does it do?

server/config.js
```diff
@@ -129,6 +129,8 @@ module.exports = {
     },
     endpoints: {
       bookings: `${prisonApiBaseUrl}/api/bookings`,
+      offenders: `${prisonApiBaseUrl}/api/offenders`,
```
spikeheap, line 132: Why have we chosen to abstract the API paths into config? I see we started in that direction before this PR, but it feels a bit odd to take on the extra complexity for something that isn't likely to change.

FWIW making the API base URL configurable makes perfect sense because this is likely to change across environments, this comment is just about the offenders endpoint.

server/config.js
```diff
@@ -129,6 +129,8 @@ module.exports = {
...
+      base: `${prisonApiBaseUrl}/api`,
```
spikeheap, line 133: I'm a bit confused about this config entry. It's composed from `prisonApiBaseUrl` but is called `base`. Which is the base?

Exposing `prisonApiBaseUrl` is fine here 👍, but we need consistency 🙂. This _could_ be an indicator that `prisonApiBaseUrl` should contain `/api` if all the examples of it append that. Otherwise it's probably safer to just type `/api` where you need it.

server/repositories/offender.js
```diff
@@ -74,6 +90,10 @@ function offenderRepository(httpClient) {
     return httpClient.get(`${endpoint}?${query.join('&')}`);
   }
 
+  function getPrisonDetailsFor(prisonId) {
+    return httpClient.get(`${getBaseUrlFrom(config)}/agencies/${[prisonId]}`);
+  }
+
```
spikeheap, line 96: This looks like it's inconsistent with the config-tree approach for generating URLs (hence why it relies on a _new_ `base` variant). That's maybe not the end of the world because we may want to rethink that approach (see another comment in this PR) – I actually favour this approach because it's clearer what your API call is than the other examples.

spikeheap, line 96: Should this live in the offender repository? This doesn't match the API structure, and feels unrelated to the other functions in this file. Do we have a `prisons` repository?

cypress/features/transactions.feature
```diff
@@ -0,0 +1,16 @@
...
+  Background: I am signed in and on the money page
+    Given that I am signed in on the money page
+
+  Scenario: I want to navigate to the Transactions page
+    When I click the "view your transactions" link
+    Then I am on the "Transactions" page
+
+
+
+
```
spikeheap, line 16: Did we agree to remove these and use vanilla cypress for now (applies to the money feature above as well)? 

I think our end-to-end tests need to be explicitly testing a valuable thing for the user. Because they're quite slow (relative to unit/integration tests) we'll end up with fewer of them, so each one wants to be carrying its own weight. 

It might just be a side-effect of trialling gherkin/cucumber here, but I think this could be a much richer test, for example checking that the correct transactions were shown in the correct order. This will be _much_ easier with the month selection addition that @LucasCairns demoed today – a single user journey can exercise the being on the page, the initial transaction set and the functionality of the date picker. So maybe don't worry about coming up with a great test for this PR.

spikeheap (changes requested):
This looks like it's pretty close, and I like the overall approach (plus, the page looks good 👍). I've added a few comments around a couple of themes:

- It looks like we might be splitting out functions into too-fine-grained decompositions, and splitting those across a repository _and_ service pattern. It'd be good to chat about the rationale in case I'm missing something, but it looks like you'd always use the service to call the repository, which indicates they're tightly coupled and maybe better encapsulated as a single thing.

- It's great to see the happy path tests 🙌. I think we need more testing of failure modes (e.g. empty responses, timeouts), exception handling and edge-cases to make this code robust across the board. I've also added some inline comments about splitting expectations out to one-per-`it`, which it might be worth us discussing to make sure it fits with how we want the tests to look.

I'll also :+1: the integration tests. Because we want to be a little sparing with our use of end-to-end tests, we need a layer of tests that check interactions between units of code. There are a couple of ways we can do that:

1. Integration test single interactions between pairs of units, mocking out everything else.
2. Integration test at (e.g.) the controller level, only mocking when we get to an API boundary.

I'd recommend option 2, having been bitten by mountains of code with the former before. Do we use `nock` or anything similar to mock out the HTTP layer in our existing test suite? Maybe this is something to explore on a call?

Anyway, there's a lot of text here, and a fair few comments, but overall this is :+1: and it's great that we're getting to explore these extra levels of testing that we weren't in a position to a year (or so) ago. 

Let's jump on a call and work through them tomorrow morning.

swilson-mojhub:
So looking for a combination of narrow integration tests (I think Wiremock is the strategic tooling of choice) and contract test. (I believe Pact is becoming popular in the teams)

server/config.js
swilson-mojhub: If I remember I _think_ this might be leftover from when we changed some of the APIs from the OG hub.

server/utils/__tests__/string.spec.js
```diff
@@ -45,3 +49,29 @@ describe('StringUtils', () => {
...
+    expect(formatted2).toBe('foo BAR');
+    const formatted3 = formatPrisonName('foo BAR baz');
+    expect(formatted3).toBe('foo BAR baz');
+  });
+
+  it('handle and empty string', () => {
+    const formatted = formatPrisonName('');
+    expect(formatted).toBe('');
+  });
+});
```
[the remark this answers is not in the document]
swilson-mojhub: Yes please.

server/services/offender/index.js
```diff
@@ -73,6 +75,34 @@ const createOffenderService = (
...
+      logger.error(
+        `OffenderService (getTransactionsFor) - Failed: ${e.message} - User: ${user.prisonerId}`,
+      );
+      logger.debug(e.stack);
+      return {
+        error: 'We are not able to show your transactions at this time',
+      };
+    }
+  }
+
```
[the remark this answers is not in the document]
LucasCairns: Refactored this code to ensure correct separation of concerns (see PrisonerInformationService)
Moved the error handling up in to the repository tier

server/services/offender/responses/prison.js
```diff
@@ -0,0 +1,26 @@
...
+    this.longDescription = longDescription;
+  }
+
+  format() {
+    return {
+      prisonId: this.prisonId,
+      description: this.description,
+      longDescription: formatPrisonName(this.longDescription),
+    };
+  }
```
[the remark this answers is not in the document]
LucasCairns: Removed the mappers entirely, in favour of just formatting just before passing to the template

server/utils/__tests__/string.spec.js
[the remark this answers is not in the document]
LucasCairns: Increased the code coverage here to ensure we assert correct behaviours for `null` or `undefined` values

server/config.js
LucasCairns: This has been removed, and for the refactored code we have decided to avoid centralising API paths in the config in favour of picking up the environment variable directly in the constructor (see PrisonApiRepository)

server/config.js
LucasCairns: For refactored code we just use the API base path in order to be more explicit with which endpoint we're calling in the repository

server/repositories/offender.js
LucasCairns: This has been removed, and for the refactored code we have decided to avoid centralising API paths in the config in favour of picking up the environment variable directly in the constructor (see PrisonApiRepository)

LucasCairns: Correct! I've decided to refactor this in to the PrisonApiRepository to be clearer with our intent

cypress/features/transactions.feature
LucasCairns: Gone! 💥

server/utils/string.js
```diff
@@ -1,5 +1,15 @@
...
 
+const formatPrisonName = description => {
+  if (!description.match(/^(HMP|HMYOI) (.+)$/)) {
+    return description;
+  }
+  return description
+    .split(' ')
+    .map((word, pos) => (pos > 0 ? capitalize(word) : word))
+    .join(' ');
+};
```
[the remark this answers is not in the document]
LucasCairns: IIRC the reasoning for this is that only intent to leave the initial word entirely uppercase and the remainder in titlecase

server/repositories/__tests__/prisonApi.spec.js
```diff
@@ -0,0 +1,221 @@
...
+          'spends',
+          new Date('2020-01-01'),
+          new Date('2020-01-31'),
+        );
+      } catch (e) {
+        hasThrown = true;
+        expect(client.get).not.toHaveBeenCalled();
+      }
+
+      expect(hasThrown).toBe(true);
```
andrewrlee, line 55: Quite a few of the other projects use `nock` so they can test the client calls are made correctly. 
There's also pretty good support in jest for async testing

An example for combining these might be:

```js
it('Should log 404 correctly', async () => {
   nock(hostname).get('/api/users/me').reply(404)

   await expect(client.get({}, '/api/users/me')).rejects.toThrow('Not Found')

   expect(logger.warn).toHaveBeenCalledWith('GET /api/users/me No record found')
 })
```

server/repositories/__tests__/prisonApi.spec.js
```diff
@@ -0,0 +1,221 @@
...
+
+  describe('getPrisonDetailsFor', () => {
+    it('should return when the prison details request succeeds', async () => {
+      const repository = new PrisonApiRepository({ client, apiUrl });
+
+      client.get.mockImplementation(() => Promise.resolve('API_RESPONSE'));
+
+      const response = await repository.getPrisonDetailsFor('TST');
+
+      expect(lastCall(client.get)[0]).toContain('/agencies/TST');
```
andrewrlee, line 148: I'm not sure what these utilities are doing, could you not just used something like? 

`expect(client.get).toHaveBeenCalledWith('/agencies/TST')`

server/repositories/__tests__/prisonApi.spec.js
```diff
@@ -0,0 +1,221 @@
...
+
+      expect(response).toBe('API_RESPONSE');
+    });
+
+    it('should throw when no prisonId is passed', async () => {
+      const repository = new PrisonApiRepository({ client, apiUrl });
+
+      client.get.mockImplementation(() => Promise.resolve('API_RESPONSE'));
```
andrewrlee, line 156: meh... you can mock returning promises using:

client.get.mockResolvedValue('API_RESPONSE');

server/repositories/__tests__/prisonApi.spec.js
```diff
@@ -0,0 +1,221 @@
...
+        expect(client.get).not.toHaveBeenCalled();
+      }
+
+      expect(hasThrown).toBe(true);
+    });
+
+    it('should swallow the error and return nothing when the request fails', async () => {
+      const repository = new PrisonApiRepository({ client, apiUrl });
+
+      client.get.mockImplementation(() => Promise.rejects('💥'));
```
andrewrlee, line 214: and similarly: `client.get.mockRejectedValue`

server/repositories/prisonApi.js
```diff
@@ -0,0 +1,93 @@
...
+const { formatISO } = require('date-fns');
+const { logger } = require('../utils/logger');
+const { getEnv } = require('../../utils');
+
+class PrisonApiRepository {
+  constructor({ client, apiUrl }) {
+    this.client = client;
+    this.url =
+      apiUrl ||
+      getEnv('PRISON_API_BASE_URL', 'https://api.nomis', {
```
andrewrlee, line 11: do you need this defaulting using env vars or do you always pass a value?

server/repositories/prisonApi.js
```diff
@@ -0,0 +1,93 @@
...
+        `PrisonApiRepository (getPrisonDetailsFor) - Failed: ${e.message} - Prison ID: ${prisonId}`,
+      );
+      logger.debug(e.stack);
+      return null;
+    }
+  }
+
+  async getBalancesFor(bookingId) {
+    if (!bookingId) {
+      throw new Error('Incorrect parameters passed');
```
andrewrlee, line 72: Never used it, but there seems to be inbuilt functionality for assertions built into node for these preconditions: https://nodejs.org/api/assert.html#assert_assert_value_message

(We tend to not use many preconditions but horses for courses)

server/routes/__tests__/money.spec.js
```diff
@@ -0,0 +1,327 @@
...
+
+  it('should prompt to sign in when the user is signed out', async () => {
+    app.use('/money', moneyRouter);
+
+    await request(app)
+      .get('/money/transactions')
+      .expect(200)
+      .then(response => {
+        const $ = cheerio.load(response.text);
+        expect($('.govuk-body').text()).toMatch(/sign in/im);
```
andrewrlee, line 86: nice!, we do really rubbish text presence check and have kept meaning to look for something like this

server/routes/money.js
```diff
@@ -1,7 +1,46 @@
 const { path } = require('ramda');
 const express = require('express');
+const { subDays } = require('date-fns');
+const { formatBalanceOrDefault, formatPrisonName } = require('../utils/string');
+const { formatDateOrDefault } = require('../utils/date');
 
-const createMoneyRouter = ({ hubContentService, offenderService }) => {
+function formatTransaction(t) {
```
andrewrlee, line 7: could possibly extract a view kind of object that creates your `data` structure and remove a lot of presentation logic out of here

server/routes/money.js
```diff
@@ -38,6 +77,65 @@ const createMoneyRouter = ({ hubContentService, offenderService }) => {
...
+      }
+
+      return res.render('pages/transactions', {
+        title: 'Your transactions',
+        config,
+        data,
+      });
+    } catch (exp) {
+      return next(exp);
+    }
```
andrewrlee, line 136: meh. We tend to wrap routes in async middleware so we don't have to remember to wrap all of our functions in try/catch... of course we then have to remember to adding the async middleware [everywhere](https://github.com/ministryofjustice/use-of-force/blob/main/server/routes/creatingReports/index.ts#L24)!

server/services/__tests__/prisonerInformation.spec.js
```diff
@@ -0,0 +1,420 @@
...
+    prisonApiRepository = {};
+  });
+
+  describe('getTransactionInformationFor', () => {
+    it('returns transaction data', async () => {
+      const prisonerInformationService = new PrisonerInformationService({
+        prisonApiRepository,
+      });
+
+      prisonApiRepository.getTransactionsFor = jest.fn(() =>
```
andrewrlee, line 68: Cos this is a class you could get jest to mock it out for you. 

```js
const PrisonApiRepository = require('some-path/prisonApiRepository')
jest.mock('some-path/prisonApiRepository')

...

const prisonApiRepository = new PrisonApiRepository(null)
prisonApiRepository.getPrisonDetailsFor.mockResolvedValue(prison)

```



```

File: server/services/prisonerInformation.js [hunk 23]
```diff
@@ -0,0 +1,84 @@
...
+      const balances = balancesResponse || {
+        error: 'We are not able to show your balances at this time',
+      };
+
+      if (transactionsResponse) {
+        const listOfPrisons = Array.from(
+          new Set(transactionsResponse.map(t => t.agencyId)),
+        );
+        const prisonDetailsResponse = await Promise.all(
+          listOfPrisons.map(p => this.prisonApi.getPrisonDetailsFor(p)),
```
andrewrlee, line 51: rather than call get prison details for each prison, you could retrieve all prisons in one go.

there's a couple of endpoints: 
https://api.prison.service.justice.gov.uk/swagger-ui/index.html#/agencies/getPrisonContactDetailList
https://api.prison.service.justice.gov.uk/swagger-ui/index.html#/agencies/getAgenciesByTypeUsingGET

server/services/prisonerInformation.js
```diff
@@ -0,0 +1,84 @@
...
+      return {
+        transactions: {
+          error: 'We are not able to show your transactions at this time',
+        },
+        balances,
+      };
+    } catch (e) {
+      logger.error(
+        `PrisonerInformationService (getTransactionInformationFor) - Failed: ${e.message}`,
+      );
```
andrewrlee, line 75: You seem to have this handle error/ coerce to null pattern at multiple levels at both the api and service layer. 

In this module you aren't checking for the api returning nulls anywhere so presumably this catch would just be catching null pointer errors which seems like using exceptions for controlling flow.

Would probably be better to just throw exceptions from the api layer and catch here rather than generating multiple exceptions.

Also regarding the logging error - I think it's sometimes nice to visualise error logging as an event stream - ideally creating one "log error" event for each error that the system produces. 

Obviously this requires ensuring there is enough context in that one error event for you to be able to work out whats going on.

andrewrlee: In java/spring bootapps you'd generally just throw (or add context and rethrow) errors to be handled by a centralised error handler and allow that to do the logging. We've have some success with using an express error handler  for similar purposes

server/utils/string.js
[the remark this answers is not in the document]
swilson-mojhub: So title case or Sentence case?  would it make sense to make an until function so make it clearer what the function is doing? 

If it is a title case you could use CSS on the template side eg `(text-transform: capitalize);`

server/utils/__tests__/string.spec.js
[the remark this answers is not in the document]
swilson-mojhub: I was referring to testing edge cases. :)

server/utils/string.js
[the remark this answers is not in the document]
LucasCairns: 100% on this ☝️  we could extract this in to a named function to make the intent clearer. It's for formatting the `longDescription` that comes back from the Prison API, so we want to format `HMP TEST LOCATION` to `HMP Test Location`.

server/repositories/__tests__/prisonApi.spec.js
LucasCairns: I think we went the route of mocking Axios through dependency injection here, however would be nice to eventually include to use something like Nock for mocking our responses and testing this integration. I'll card this one up as it's something I would would prefer to introduce separately than quickly fit in this PR

server/repositories/__tests__/prisonApi.spec.js
LucasCairns: Totally agree here, prefer to use the Jest in-builts when possible. Updated the tests

server/repositories/__tests__/prisonApi.spec.js
LucasCairns: Totally agree here, prefer to use the Jest in-builts when possible. Updated the tests

server/repositories/__tests__/prisonApi.spec.js
LucasCairns: Totally agree here, prefer to use the Jest in-builts when possible. Updated the tests

server/repositories/prisonApi.js
LucasCairns: I've updated these functions to check using Assert the validity of the arguments passed before making a call to the API. Something I hope to remove later down the line if we move to TS 🤔

server/utils/string.js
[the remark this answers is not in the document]
LucasCairns: This code has now been removed as we no longer need to format the prison name. We switched to another endpoint that provides a formatted name for us, consistent with staff facing services.

server/routes/__tests__/money.spec.js
LucasCairns: 😄

server/routes/money.js
LucasCairns: Love the consistency/readability the wrapper functions adds ❤️ Carding this up as technical debt, I've bookmarked that link 👍

server/services/__tests__/prisonerInformation.spec.js
LucasCairns: Arrgh.. frustratingly it doesn't support named exports (something we prefer and use all over the place), without defining our own mock implementation, which is what we already have 😞

server/services/prisonerInformation.js
LucasCairns: Ooooh good shout! I've updated our call to use the above endpoints, cheers! 👍

server/services/prisonerInformation.js
LucasCairns: My intent with swallowing the errors in the API tier was to allow for some graceful degradation when we unable to surface all the information for the page, such that if a call for their balance succeeds and the transactions call fails, then we would just show them their balance and notify them to try again for their transactions.

Interested in the logging by the way 🤔 would love to see some examples if you have them at hand

server/routes/money.js
LucasCairns: This has been removed following review from Andy bellow, it was to get a unique list of prisons from the transactions in order to query the Prison API, it was no longer required following the API change

server/services/offender/responses/transaction.js
LucasCairns: 100% 👍 Updated to retain the pence amount until we're ready to present the data in the template

server/services/offender/responses/transaction.js
```diff
@@ -0,0 +1,56 @@
...
+          : null,
+      moneyOut:
+        this.postingType === 'DR'
+          ? formatBalanceOrDefault(null, this.amount, this.currency)
+          : null,
+      balance: formatBalanceOrDefault(DEFAULT, this.balance, this.currency),
+      paymentDescription: this.paymentDescription,
+      prison: this.prison,
+    };
+  }
```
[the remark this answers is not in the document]
LucasCairns: Refactored and separated this out so there is now a mapper for formatting this data on the controller before passing to the template

server/services/offender/responses/__tests__/prison.js
```diff
@@ -0,0 +1,17 @@
...
+      description: 'Test Prison (HMP)',
+      longDescription: 'HMP TEST PRISON',
+    };
+
+    const formatted = Prison.from(response).format();
+
+    expect(formatted.prisonId).toBe('TST');
+    expect(formatted.description).toBe('Test Prison (HMP)');
+    expect(formatted.longDescription).toBe('HMP Test Prison');
+  });
```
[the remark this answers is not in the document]
LucasCairns: The code this is testing has been removed

server/utils/string.js
```diff
@@ -1,5 +1,15 @@
 const { capitalize } = require('./index');
 
+const formatPrisonName = description => {
+  if (!description.match(/^(HMP|HMYOI) (.+)$/)) {
+    return description;
+  }
```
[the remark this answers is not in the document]
LucasCairns: Removed, this is no longer required since we now get the formatted location from the Prison API

server/utils/__tests__/string.spec.js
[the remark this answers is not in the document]
LucasCairns: Removed, this is no longer required since we now get the formatted location from the Prison API

server/repositories/prisonApi.js
LucasCairns: I've just updated this to require the environment variable to be set, otherwise we throw. I think we should be explicit in our configuration to avoid any mishaps 🤔

spikeheap (approved):
:+1: this is looking good. 🙌 for addressing all the rounds of changes 😍

spikeheap (approved):

PR #209, opened 2021-03-10
🧹 Remove NPM verify from Dockerfile, tests are run in CI

spikeheap:
### Context

No

### Intent

This PR removes the test step from the Dockerfile, because we run tests as part of CI.


davidthomas22 (approved):
LGTM

LucasCairns (approved):
👍

PR #226, opened 2021-03-29
Feature/pending transactions

LucasCairns:
### Context

https://trello.com/c/ohGgiiSP

### Intent

- Add pending transactions to when viewing the Private(cash) account

Screenshot 2021-03-29 at 17 59 35
Screenshot 2021-03-29 at 18 00 03

### Considerations

- Generation of user notification has been moved from the service to the presentation layer
- Refactored the naming of things in `formatters.js` to be clearer with the intent
- Added additional coverage with Sentry when processing response data

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/routes/money.js
+++ b/server/routes/money.js
@@ -2,8 +2,9 @@ const { path } = require('ramda');
 const express = require('express');
 const { startOfMonth, parseISO, endOfMonth, isFuture } = require('date-fns');
 const {
-  formatTransactionPageData,
-  formatDamageObligations,
+  createTransactionsResponseFrom,
+  createDamageObligationsResponseFrom,
+  createPendingTransactionsResponseFrom,
 } = require('./formatters');
 const { getDateSelection } = require('../utils/date');
 
@@ -79,7 +80,9 @@ const createMoneyRouter = ({
         );
 
         templateData.prisonerInformation = {
-          damageObligations: formatDamageObligations(damageObligations),
+          damageObligations: createDamageObligationsResponseFrom(
+            damageObligations,
+          ),
           selected: 'damage-obligations',
           accountTypes: ['spends', 'private', 'savings'],
         };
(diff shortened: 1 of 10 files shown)


server/views/components/personal-transactions/template.njk
```diff
@@ -29,107 +27,144 @@
         </a>
       </li>
     {% endfor %}
-    {% if (data.selected === 'damage-obligations') or (data.shouldShowDamageObligationsTab) %}
+    {% if (data.selected === 'damage-obligations') or data.shouldShowDamageObligationsTab %}
```
andrewrlee, line 30: was just wondering whether you had considered modelling this as 3 different routes rather than hiding/showing each section using javascript? 

imagine it would have simplified the service layer and pages and you could more easily track who was accessing each which tab. 

We do something similar in use of force, would there have been any negative side effects from doing it that way?

server/views/components/personal-transactions/template.njk
```diff
@@ -29,107 +27,144 @@
...
+    {% if data.pendingTransactions and data.pendingTransactions.userNotification %}
+      <p class="govuk-!-font-size-24">
+        {{data.pendingTransactions.userNotification}}, <a href="{{ returnUrl }}" class="govuk-link">try again</a>
+      </p>
+    {% elif data.pendingTransactions.rows %}
+      <h2 class="govuk-heading-m govuk-!-margin-bottom-1">Pending payments</h2>
+      <p>These are payments you have received but have not yet cleared. Pending payments do not show in your current balance</p>
+      <table id="pending-transactions" class="govuk-table">
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
```
andrewrlee, line 66: the table macro would have been quite nice for this (unless there's something weird going on which means you couldn't have used it)

davidthomas22 (approved):
LGTM

server/views/components/personal-transactions/template.njk
LucasCairns: Probably not, I think it's just evolved in to this as we've worked on it. 100% on splitting things out, I've just thrown a commit up that splits out the routes and templates to reduce the amount of conditional logic

server/views/components/personal-transactions/template.njk
LucasCairns: I don't there was an explicit reason, so I've updated the code to use GovUK tables instead

server/routes/money.js
```diff
@@ -155,6 +239,72 @@ const createMoneyRouter = ({
...
+          header: false,
+          postscript: true,
+          detailsType: 'small',
+          returnUrl: req.originalUrl,
+        },
+      };
+
+      const { user } = req;
+
+      if (user) {
```
andrewrlee, line 257: Not a specific query for this PR but does this only happen when a user is logged out? 

I'm obviously missing a lot of context but it seems like some pages are decorated with additional info when a user is logged in but are still accessible when they aren't. 

Does it make sense that a user can access this page when not logged in? 
Also are links to this page available when a user is not logged in?

server/routes/money.js
```diff
@@ -155,6 +239,72 @@ const createMoneyRouter = ({
...
+        );
+
+        const transactionsData = await prisonerInformationService.getPrivateTransactionsFor(
+          user,
+          fromDate,
+          toDate,
+        );
+
+        if (!transactionsData) {
+          return next('Failed to fetch transaction data');
```
andrewrlee, line 270: does the error handler support a string error?

LucasCairns: Good point, no it does not. Fixed

server/routes/money.js
LucasCairns: Probably not for this page, but historically for personal information;
- Signing in has always been optional, prompts to sign in have been on the personal information components themselves
- The user is able to sign out on these pages and we wanted to preserve the their history so avoided redirects
- We've displayed personal information alongside related content, so it made sense to still display the page

But 100% agree this boilerplate is some-what redundant here. I think - and probably not for this PR, it would be worth exploring a more elegant approach with middleware down the line

andrewrlee (approved):
LGTM

PR #229, opened 2021-04-07
A few small changes:

andrewrlee:
### Context

No, this just covers a few documentation fixes and a logging fix

### Intent

 * Removing unused env vars and incorrect endpoint url from env template
 * Updating doc'd node version
 * Only log missing hot jar at startup rather than every request

No

### Considerations

No 

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -224,12 +224,12 @@ const createApp = ({
   }
 
   // Routing
+  if (!process.env.HOTJAR_ID) {
+    logger.warn('HOTJAR_ID not set');
+  }
 
   app.use((req, res, next) => {
     const hotJarId = process.env.HOTJAR_ID;
-    if (!hotJarId) {
-      logger.warn('HOTJAR_ID not set');
-    }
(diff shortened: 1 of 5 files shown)


server/app.js
```diff
@@ -224,12 +224,12 @@ const createApp = ({
   }
 
   // Routing
+  if (!process.env.HOTJAR_ID) {
```
andrewrlee, line 227: this was a bit noisy - so moved to just on start up

spikeheap: 🙌 would it be worth sending this warning to Sentry as well?

spikeheap (dismissed):

davidthomas22 (dismissed):
LGTM

server/app.js
andrewrlee: added capturing a message in sentry. 

Docs say that these are not usually emitted. not sure if it's specific config we need to enable

davidthomas22 (approved):
LGTM

PR #230, opened 2021-04-07
Allow retrieval of the names of inactive prisons 

andrewrlee:
### Context

https://trello.com/c/O82U3mp8

View a transaction that occurred in an inactive prison such as Camp Hill

### Intent

Altering the code that performs the prison name lookup to include inactive prisons. 

### Considerations

If I generate two lists of prisons from both the old and new endpoints, it produces the exact same output
```
curl -X GET  -v  "https://api-preprod.prison.service.justice.gov.uk/api/agencies/type/INST?activeOnly=true" -H "Authorization: Bearer $TOKEN" | jq '.[] | .description' > insts.txt

curl -X GET  -v  "https://api-preprod.prison.service.justice.gov.uk/api/agencies/prison" -H "Authorization: Bearer $TOKEN" | jq '.[] | .formattedDescription' > prisons.txt 
```

If I then switch the `activeOnly` flag to `false` THE DIFF shows that the inactive prisons are added. 

This API method is only used to look up prison names for a specific agency

### Checklist

- [x] Tested in Development

### DIFF
--- a/server/routes/__tests__/money.spec.js
+++ b/server/routes/__tests__/money.spec.js
@@ -78,8 +78,7 @@ describe('Prisoner Money', () => {
   const agencyApiResponse = [
     {
       agencyId: 'TST',
-      description: 'TEST (HMP)',
-      formattedDescription: 'Test (HMP)',
+      description: 'Test (HMP)',
     },
   ];
 
@@ -154,7 +153,7 @@ describe('Prisoner Money', () => {
         if (requestUrl.match(/\/transaction-history/i)) {
           return Promise.resolve(transactionApiResponse);
         }
-        if (requestUrl.match(/\/agencies\/prison/i)) {
+        if (requestUrl.match(/\/agencies\/type\/INST\?activeOnly=false/i)) {
           return Promise.resolve(agencyApiResponse);
         }
         if (requestUrl.match(/\/balances/i)) {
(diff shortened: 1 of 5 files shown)


davidthomas22 (approved):
LGTM

server/services/prisonerInformation.js
```diff
@@ -4,7 +4,7 @@ const { logger } = require('../utils/logger');
 
 function createPrisonFormatter(listOfPrisons, key) {
   const lookup = listOfPrisons.reduce((a, p) => {
-    a.set(p.agencyId, p.formattedDescription);
+    a.set(p.agencyId, p.description);
```
jonhindle, line 7: Coding style query here really. Favour destructuring? Auto return single line? I should probably check our linting preferences too actually 🤔

`listOfPrisons.reduce((a, { agencyId, description } ) => 
    a.set(agencyId, description), newMap());`
 
...also I think you _could_ possibly make it clearer we're returning a Map (I think this is reasonably widely supported)...

`new Map(listOfPrisons.map(({ agencyId, description }) => [agencyId, description]));`

server/routes/__tests__/money.spec.js
```diff
@@ -154,7 +153,7 @@ describe('Prisoner Money', () => {
         if (requestUrl.match(/\/transaction-history/i)) {
           return Promise.resolve(transactionApiResponse);
         }
-        if (requestUrl.match(/\/agencies\/prison/i)) {
+        if (requestUrl.match(/\/agencies\/type\/INST\?activeOnly=false/i)) {
```
jonhindle, line 156: Inherited I appreciate, but the replication of the regexp, kinda suggests that it might be better as a const or wrapped in a descriptive function 🤷‍♀️ Probably a code style thing...

server/services/prisonerInformation.js
andrewrlee: Didn't know that set returned the map!
I like the second example too

server/routes/__tests__/money.spec.js
andrewrlee: this needs tidying up generally

PR #232, opened 2021-04-12
Feature/add timetable functionality profile page

davidthomas22:
### Context

### Intent

### Considerations

### Checklist

- [ ] Tested in Development


server/__tests__/app.spec.js
```diff
@@ -129,7 +129,7 @@ describe('App', () => {
           hubFeaturedContent: jest.fn().mockRejectedValue(error),
         },
         offenderService: {
-          getEventsForToday: jest.fn().mockResolvedValue([]),
+          getEventsListForToday: jest.fn().mockResolvedValue([]),
```
andrewrlee, line 132: rename - this returns a list of events for the home page, where as we wanted to add a `getEventsForToday` that returned the same type as `getEventsFor`

server/routes/profile.js
```diff
@@ -1,25 +1,36 @@
 const express = require('express');
 
-const createProfileRouter = () => {
+const createProfileRouter = ({ offenderService }) => {
   const router = express.Router();
 
+  const getPersonalisation = async user => {
+    const events = await offenderService.getEventsForToday(user);
+
+    const timetableError = events?.error;
```
andrewrlee, line 9: events can never be null so won't need the `?.`

As Jon suggested this could probably be moved into the line below: 
`const { morning, afternoon, evening, error: timeTableError } = events || {};`

server/routes/profile.js
```diff
@@ -1,25 +1,36 @@
...
+    const { morning, afternoon, evening } = events || {};
+    const signedInUser = user.getFullName();
```
andrewrlee, line 11: this signedInUser should probably move into middleware at some point and added to res.locals.

server/routes/profile.js
```diff
@@ -1,25 +1,36 @@
...
+      signedInUser,
+      events: { error: timetableError, morning, afternoon, evening },
+    };
+  };
+
   router.get('/', async (req, res, next) => {
     try {
-      const templateParams = {
+      const { user } = req;
+      const personalisation = user ? await getPersonalisation(user) : {};
```
andrewrlee, line 22: Not one for now but worth contemplating, we could potentially have some snazzy personalisation middleware to decouple the routing code and the personalisation code.

server/services/__tests__/offenderService.spec.js
```diff
@@ -205,14 +206,54 @@ describe('Offender Service', () => {
       });
 
       const data = await service.getEventsFor(
-        TEST_BOOKING_ID,
+        TEST_USER,
```
andrewrlee, line 209: this test was always failing for the wrong reason as a number was being passed instead of a test user.

server/services/offender/index.js
```diff
@@ -215,7 +215,7 @@ const createOffenderService = (
         throw new Error('Invalid dates supplied');
       }
 
-      if (!isBefore(startDateObj, endDateObj)) {
+      if (isBefore(endDateObj, startDateObj)) {
```
andrewrlee, line 218: fix this to allow creating events for today

andrewrlee: noticed that not on events in this service use Sentry... think it would be better to take a step back and look at better patterns for error handling than fix as this PR.

server/routes/profile.js
andrewrlee: this is now addressed

andrewrlee (approved):
lgtm

assets/javascript/card.js
```diff
@@ -0,0 +1,12 @@
...
+  document.querySelectorAll('.card--clickable').forEach(card => {
+    // Check if card has a link within it
+    if (card.querySelector('a') !== null) {
+      // Clicks the link within the heading to navigate to desired page
+      card.addEventListener('click', () => {
+        card.querySelector('a').click();
+      });
+    }
+  });
+};
```
andrewrlee, line 12: stolen directly from prison staff hub

assets/sass/components/_card.scss
```diff
@@ -0,0 +1,100 @@
+/* ==========================================================================
```
andrewrlee, line 1: stolen directly from prison staff hub

PR #250, opened 2021-05-06
Extracting out a topic service and new json api client

andrewrlee:
### Context

https://trello.com/c/cuCgbKtD

### Intent

Put in place the start of a new json api client and extract out a cmsApi which will be the basis of new json api functionality.

I've then moved the existing json api call into it and ensured its covered by nock tests

As part of this I realised that the sorting on the topic page only checked the first character. This lead to strange results in certain cases: 
Screenshot 2021-05-05 at 15 37 53

### Considerations

This has changed some of the error handling behaviour. 

The topic page is made out of landing pages and categories - previously if either call failed then it would still show the results of the other call. Not if one call fails it will show an error page. 

Could look at reinstating this if necessary, but going forward we will hopefully merge both calls into one by making changes in drupal.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,16 +31,6 @@
     "node": ">= 14",
     "npm": ">=6.14.8"
   },
-  "nodemonConfig": {
-    "ignore": [
-      ".circleci/*",
-      "migrations/*",
-      "node_modules/*",
-      "**/__tests__/**",
-      "build-info.json"
-    ],
-    "delay": "2500"
-  },
   "jest": {
     "verbose": true,
     "testMatch": ["**/__tests__/**/*.[jt]s?(x)"]
--- a/server/app.js
+++ b/server/app.js
@@ -42,7 +42,7 @@ const createApp = ({
   logger,
   requestLogger,
   hubFeaturedContentService,
-  hubMenuService,
+  topicsService,
   hubContentService,
   hubTagsService,
   offenderService,
@@ -247,7 +247,7 @@ const createApp = ({
     }),
   );
 
-  app.use('/topics', createTopicsRouter({ hubMenuService }));
+  app.use('/topics', createTopicsRouter({ topicsService }));
 
   app.use('/timetable', createTimetableRouter({ offenderService }));
 
--- a/server/clients/jsonApiClient.js
+++ b/server/clients/jsonApiClient.js
@@ -1,0 +1,35 @@
+const axios = require('axios');
+const httpAdapter = require('axios/lib/adapters/http');
+const { logger } = require('../utils/logger');
+
+class JsonApiClient {
+  constructor(baseURL) {
+    this.client = axios.create({
+      baseURL,
+      timeout: 30000,
+      adapter: httpAdapter,
+    });
+    this.client.interceptors.request.use(request => {
+      logger.info(
+        `JsonApiClient [${request.method}] request - ${request.baseURL}${request.url}`,
+      );
+      return request;
+    });
+
+    this.client.interceptors.response.use(res => {
+      logger.info(
+        `JsonApiClient [${res.config?.method}], status: ${res.status} response - ${res.config?.baseURL}${res.config?.url}`,
+      );
+      return res;
+    });
+  }
+
+  async get(path, { query } = {}) {
+    const res = await this.client.get(path, { params: query });
+    return res.data;
+  }
+}
+
+module.exports = {
+  JsonApiClient,
+};
--- a/server/config.js
+++ b/server/config.js
@@ -41,13 +41,13 @@ module.exports = {
     callbackPath: '/auth/provider/callback',
   },
   api: {
+    hubEndpoint,
     hubContent: `${hubEndpoint}/v1/api/content`,
     hubCategoryFeatured: `${hubEndpoint}/v1/api/category/featured`,
     categoryMenu: `${hubEndpoint}/v1/api/category-menu`,
     hubTerm: `${hubEndpoint}/v1/api/term`,
     series: `${hubEndpoint}/v1/api/content/series`,
     tags: `${hubEndpoint}/v1/api/vocabulary/tags`,
-    primary: `${hubEndpoint}/jsonapi/node/landing_page?fields[node--landing_page]=title,field_moj_description,drupal_internal__nid,field_moj_prisons`,
   },
   apiV2: {
     hubContentFeatured: `${hubEndpoint}/v2/api/content/featured`,
--- a/server/repositories/__tests__/cmsApi.spec.js
+++ b/server/repositories/__tests__/cmsApi.spec.js
@@ -1,0 +1,132 @@
+const nock = require('nock');
+const {
+  primaryMenuResponse,
+  prisons,
+  landingPage,
+} = require('../../../test/resources/primaryMenu');
+const { JsonApiClient } = require('../../clients/jsonApiClient');
+const { CmsApi } = require('../cmsApi');
+
+describe('CmsApi', () => {
+  let cmsApi;
+  let mockDrupal;
+
+  beforeEach(() => {
+    if (!nock.isActive()) {
+      nock.activate();
+    }
+    const host = 'http://localhost:3333';
+    cmsApi = new CmsApi(new JsonApiClient(host));
+    mockDrupal = nock(host);
+  });
+
+  afterEach(() => {
+    nock.restore();
+  });
+
+  describe('primaryMenu', () => {
+    const path =
+      '/jsonapi/node/landing_page?fields[node--landing_page]=title,field_moj_description,drupal_internal__nid,field_moj_prisons';
+
+    it('should handle no items', async () => {
+      mockDrupal.get(path).reply(200, primaryMenuResponse([]));
+
+      const response = await cmsApi.primaryMenu(792);
+
+      expect(response).toStrictEqual([]);
+    });
+
+    it('should handle empty responses', async () => {
+      mockDrupal.get(path).reply(200);
+
+      const response = await cmsApi.primaryMenu(792);
+
+      expect(response).toStrictEqual([]);
+    });
+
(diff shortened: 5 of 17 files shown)


package.json
```diff
@@ -31,16 +31,6 @@
     "node": ">= 14",
     "npm": ">=6.14.8"
   },
-  "nodemonConfig": {
```
andrewrlee, line 34: this was being overriden by the nodemon.json file so have removed this and updated the json config file.

server/app.js
```diff
@@ -42,7 +42,7 @@ const createApp = ({
   logger,
   requestLogger,
   hubFeaturedContentService,
-  hubMenuService,
+  topicsService,
```
andrewrlee, line 45: renamed as this is all this service actually does

server/clients/jsonApiClient.js
```diff
@@ -0,0 +1,35 @@
+const axios = require('axios');
+const httpAdapter = require('axios/lib/adapters/http');
+const { logger } = require('../utils/logger');
+
+class JsonApiClient {
+  constructor(baseURL) {
+    this.client = axios.create({
+      baseURL,
+      timeout: 30000,
```
andrewrlee, line 9: hardcoded to a relatively large timeout - before there was no timeout and calls could potentially block forever.
We could look at extracting this out into config..

server/clients/jsonApiClient.js
```diff
@@ -0,0 +1,35 @@
+const axios = require('axios');
+const httpAdapter = require('axios/lib/adapters/http');
+const { logger } = require('../utils/logger');
+
+class JsonApiClient {
+  constructor(baseURL) {
```
andrewrlee, line 6: All the other base clients we have require the verb methods to specify the url. This client only requires the path specifying, so:  

`jsonApi.get('/some-path')`

vs

`jsonApi.get('http:some-site/some-path')`

server/clients/jsonApiClient.js
```diff
@@ -0,0 +1,35 @@
...
+      );
+      return request;
+    });
+
+    this.client.interceptors.response.use(res => {
+      logger.info(
+        `JsonApiClient [${res.config?.method}], status: ${res.status} response - ${res.config?.baseURL}${res.config?.url}`,
+      );
+      return res;
+    });
```
andrewrlee, line 24: centralising logging in interceptors rather than for each method

server/clients/jsonApiClient.js
```diff
@@ -0,0 +1,35 @@
+const axios = require('axios');
+const httpAdapter = require('axios/lib/adapters/http');
+const { logger } = require('../utils/logger');
+
+class JsonApiClient {
+  constructor(baseURL) {
+    this.client = axios.create({
+      baseURL,
+      timeout: 30000,
+      adapter: httpAdapter,
```
andrewrlee, line 10: did this to make life a bit easier - in tests, axios detects js-dom and uses a fake browser based xmlhttprequest object which doesn't play nice with jest and is not representative as to how the client would work in production.

explicitly specifying the adapter in the prod code stops that from happening

server/config.js
```diff
@@ -41,13 +41,13 @@ module.exports = {
...
   },
   api: {
+    hubEndpoint,
     hubContent: `${hubEndpoint}/v1/api/content`,
     hubCategoryFeatured: `${hubEndpoint}/v1/api/category/featured`,
     categoryMenu: `${hubEndpoint}/v1/api/category-menu`,
     hubTerm: `${hubEndpoint}/v1/api/term`,
     series: `${hubEndpoint}/v1/api/content/series`,
     tags: `${hubEndpoint}/v1/api/vocabulary/tags`,
-    primary: `${hubEndpoint}/jsonapi/node/landing_page?fields[node--landing_page]=title,field_moj_description,drupal_internal__nid,field_moj_prisons`,
```
andrewrlee, line 50: as we replace bits with json api I'd hope that we'd slowly get rid of all of these

server/repositories/__tests__/cmsApi.spec.js
```diff
@@ -0,0 +1,132 @@
...
+
+      return expect(cmsApi.primaryMenu(792)).rejects.toEqual(
+        Error('Request failed with status code 500'),
+      );
+    });
+
+    it('should format and return matching item', async () => {
+      mockDrupal.get(path).reply(
+        200,
+        primaryMenuResponse([
```
andrewrlee, line 58: created this kind of test data builder as think it makes it a bit clearer as to what we're testing rather than pointing to random json files

server/repositories/cmsApi.js
```diff
@@ -0,0 +1,37 @@
...
+    const response = await this.jsonApiClient.get(
+      `/jsonapi/node/landing_page?fields[node--landing_page]=title,field_moj_description,drupal_internal__nid,field_moj_prisons`,
+    );
+    return this.transform(response?.data || {}, prisonId);
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  transform(data, prisonId) {
+    const items = Object.values(data)
+      .filter(({ relationships }) => {
```
andrewrlee, line 18: this filter can go once  we move to the json api endpoints that allow you to specify a prison as this bit is all handled in drupal

server/repositories/cmsApi.js
```diff
@@ -0,0 +1,37 @@
+const { getEstablishmentUiId } = require('../utils');
+
+class CmsApi {
+  constructor(jsonApiClient) {
+    this.jsonApiClient = jsonApiClient;
+  }
+
+  async primaryMenu(prisonId = 0) {
+    const response = await this.jsonApiClient.get(
+      `/jsonapi/node/landing_page?fields[node--landing_page]=title,field_moj_description,drupal_internal__nid,field_moj_prisons`,
```
andrewrlee, line 10: will look at replacing this using drupal query string building  library in a follow up

server/repositories/hubMenu.js
```diff
@@ -1,22 +1,13 @@
-const { path } = require('ramda');
 const config = require('../config');
 const {
   tagIdFrom,
   nameFrom,
   termDescriptionValueFrom,
 } = require('../selectors/hub');
-const { getEstablishmentUiId } = require('../utils');
 
-function hubMenuRepository(httpClient, jsonClient) {
```
andrewrlee, line 10: this only sorted by first character

server/routes/topics.js
```diff
@@ -37,6 +38,7 @@ const createTopicsRouter = ({ hubMenuService }) => {
         config,
       });
     } catch (e) {
+      logger.error(`Error loading topics: ${e.message}`);
```
andrewrlee, line 41: errors propagate up to here so doing the logging in one place... hopefully we could push this into a global error handler

server/services/topics.js
```diff
@@ -0,0 +1,23 @@
...
+  constructor(hubMenuRepository, cmsApi) {
+    this.hubMenuRepository = hubMenuRepository;
+    this.cmsApi = cmsApi;
+  }
+
+  async getTopics(prisonId) {
+    const [tags, primary] = await Promise.all([
+      this.hubMenuRepository.tagsMenu(prisonId),
+      this.cmsApi.primaryMenu(prisonId),
+    ]);
```
andrewrlee, line 13: now making requests in parallel and sorting by whole string rather than first letter.

server/repositories/cmsApi.js
andrewrlee: when this goes we can get rid of quite a bit of code

server/services/__tests__/topicsService.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    it('Sources to have been called correctly', async () => {
+      hubMenuRepo.tagsMenu.mockResolvedValue([]);
+      cmsApi.primaryMenu.mockResolvedValue([]);
+
+      await topicService.getTopics(1);
+
+      expect(hubMenuRepo.tagsMenu).toHaveBeenCalledWith(1);
+      expect(cmsApi.primaryMenu).toHaveBeenCalledWith(1);
+    });
```
jonhindle, line 79: Duplicate test.

server/routes/topics.js
```diff
@@ -13,14 +14,14 @@ const fixUrls = element => {
...
-const createTopicsRouter = ({ hubMenuService }) => {
+const createTopicsRouter = ({ topicsService }) => {
   const router = express.Router();
 
   router.get('/', async (req, res, next) => {
     try {
       const userName = req.user && req.user.getFullName();
       const establishmentId = path(['session', 'establishmentId'], req);
-      const topics = await hubMenuService.allTopics(establishmentId);
+      const topics = await topicsService.getTopics(establishmentId);
```
jonhindle, line 24: minor, but you could have destructed out this `({ topicsService: { getTopics } })`

andrewrlee: kinda like having the reference to the service in the code where it's used. 

Can make it easier to find out where things are being used - we have some places where similar named methods are declared at different levels. I personally find it easier to see the context of where the function is declared in the code - we have quite a few functions with the same name declared in different functions.

It also helps evaluate cohesiveness easier if you can easily see which functions in a class use specific dependencies, rather than a detached function.

server/services/__tests__/topicsService.spec.js
andrewrlee: removed this!

server/services/topics.js
```diff
@@ -0,0 +1,23 @@
+const caseInsensitive = new Intl.Collator('en', { caseFirst: 'lower' });
+
+class TopicsService {
+  constructor(hubMenuRepository, cmsApi) {
+    this.hubMenuRepository = hubMenuRepository;
+    this.cmsApi = cmsApi;
```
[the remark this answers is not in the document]
andrewrlee: Done this.. kind annoying how verbose it makes it

PR #259, opened 2021-05-14
feature/using-json-api-for-topic-list

andrewrlee:
Moving to use the json api to retrieve the topic list.

This one api returns both the categories and tags that make up the task list.

### Context

https://trello.com/c/S9qfocTO

### Intent

Moves from using two separate custom apis to one single json api.

### Considerations

The current json api seems to return tags with no content - like appeals, HDC etc. 

Need to talk to Leon about this before merging.


server/content/establishmentData.json
```diff
@@ -3,7 +3,6 @@
     "name": "berwyn",
     "formattedName": "Berwyn",
     "prefix": "HMP",
-    "uuId": "fd1e1db7-d0be-424a-a3a6-3b0f49e33293",
```
andrewrlee, line 6: these UUIDs were used to filter categories but the new api filters by prison on the drupal side

server/repositories/__tests__/cmsApi.spec.js
```diff
@@ -24,58 +24,63 @@ describe('CmsApi', () => {
...
+      );
+    });
+
     it('should format and return matching item', async () => {
       mockDrupal.get(path).reply(
         200,
-        primaryMenuResponse([
-          landingPage({
+        jsonApiResponse([
+          tag({
```
andrewrlee, line 66: Moved to use drupal domain terminology within the api/data layer: `tag` `categories` 

and exposing app domain terminology throughout the rest of the codebase: `topics`

server/repositories/cmsApi.js
```diff
@@ -1,37 +1,70 @@
...
+  ])
+  .addFilter(
+    'vid.meta.drupal_internal__target_id',
+    ['moj_categories', 'tags'],
+    'IN',
+  )
+  .addPageLimit(100)
+  .getQueryString();
 
 class CmsApi {
```
andrewrlee, line 23: This is going to grow but don't really want to presuppose a tidier structure until we have some additional queries.

Can envision potentially extracting out some query object which encapsulates the query and deserializing but probably makes sense to do this when we have more examples

server/repositories/hubMenu.js
```diff
@@ -46,7 +35,6 @@ function hubMenuRepository(httpClient) {
 
   return {
     categoryMenu,
-    tagsMenu,
```
andrewrlee, line 49: no longer used

server/routes/__tests__/health.spec.js
```diff
@@ -1,14 +1,14 @@
...
 
 const { createHealthRouter } = require('../health');
 
 describe('GET healthchecks', () => {
   let config;
   let app;
-  const BUILD_NUMBER = 'foo-number'
-  const GIT_REF = 'foo-ref'
-  const GIT_DATE = 'foo-date'
+  const BUILD_NUMBER = 'foo-number';
```
andrewrlee, line 9: having to format this as a commit managed to get in that escaped the linting check

server/repositories/cmsApi.js
```diff
@@ -1,37 +1,70 @@
...
     this.jsonApiClient = jsonApiClient;
   }
 
-  async primaryMenu(prisonId = 0) {
-    const response = await this.jsonApiClient.get(
-      `/jsonapi/node/landing_page?fields[node--landing_page]=title,field_moj_description,drupal_internal__nid,field_moj_prisons`,
+  async getTopics(establishmentName) {
+    if (!establishmentName) {
+      return [];
+    }
```
andrewrlee, line 31: think this might be a bit lame - could potentially ensure that we have a prison in context on the frontend and throw if not.

server/repositories/cmsApi.js
```diff
@@ -1,37 +1,70 @@
...
+
+    const { data } = await this.jsonApiClient.get(
+      `/jsonapi/prison/${establishmentName}/taxonomy_term?${topicsQuery}`,
     );
-    return this.transform(response?.data || {}, prisonId);
+
+    return this.transform(data || []);
```
andrewrlee, line 37: again this could be a bit lame - can't foresee why drupal would return us a falsey value... maybe should hard fail here?

server/routes/health.js
```diff
@@ -1,6 +1,6 @@
 const express = require('express');
 
-const createHealthRouter = ({ buildInfo:build }) => {
+const createHealthRouter = ({ buildInfo: build }) => {
```
andrewrlee, line 3: having to format this as a commit managed to get in that escaped the linting check

server/routes/topics.js
```diff
@@ -1,27 +1,17 @@
 const express = require('express');
-const { path } = require('ramda');
 const Sentry = require('@sentry/node');
 const { logger } = require('../utils/logger');
 
-const fixUrls = element => {
```
andrewrlee, line 6: need a sanity check on this... it wasn't doing anything right?!

server/routes/topics.js
```diff
@@ -1,27 +1,17 @@
...
   const router = express.Router();
 
   router.get('/', async (req, res, next) => {
     try {
       const userName = req.user && req.user.getFullName();
-      const establishmentId = path(['session', 'establishmentId'], req);
-      const topics = await topicsService.getTopics(establishmentId);
+      const { establishmentId, establishmentName } = req.session;
+      const topics = await topicsService.getTopics(
+        establishmentId && establishmentName,
```
andrewrlee, line 13: maybe should fix at source - it seems establishment name is always set based on dodgy regex, we only know its a valid establishment if we can find the associated id.

server/services/topics.js
```diff
@@ -1,24 +1,17 @@
...
-      this.#cmsApi.primaryMenu(prisonId),
-    ]);
-
-    return tags
-      .concat(primary)
-      .sort((a, b) => caseInsensitive.compare(a.linkText, b.linkText));
+    const topics = await this.#cmsApi.getTopics(prisonId);
+    return topics.sort((a, b) =>
+      caseInsensitive.compare(a.linkText, b.linkText),
+    );
```
andrewrlee, line 14: could potentially sort in the api layer?

server/repositories/cmsApi.js
leonk: As long as the request is considered a valid JSON:API url, then Drupal will always return `data` (and will be empty for no results).  And you will also get a 200 response code.

But there are quite a few other response types, if the url isn't valid.

- `/jsonapi/some-unknown-entity-type` will return a 404 (with an `errors` property and no `data`)
- `/jsonapi/taxonomy_term?filter[some-uknown-field]=1` will return a 400 (with an `errors` property and no `data`)
- `/jsonapi/taxonomy_term?filter=incorrect-way-to-use-a-filter` will return a 500 (this appears to be a bug in Drupal, as it's not able to deal with the request at all and results in a fatal error).

server/services/topics.js
leonk: Yeah I think ideally sorting would be done via JSON:API.  You should be able to use `?sort=name`.

server/repositories/cmsApi.js
andrewrlee: cool, I've simplified this, I think any non 2xx/3xx response will result in an exception and a user facing error being presented.

server/repositories/cmsApi.js
```diff
@@ -1,37 +1,80 @@
...
 class CmsApi {
   constructor(jsonApiClient) {
     this.jsonApiClient = jsonApiClient;
   }
 
-  async primaryMenu(prisonId = 0) {
+  async getTopics(establishmentName) {
+    if (!establishmentName) {
+      return [];
+    }
```
jonhindle, line 42: Single line:
`if (!establishmentName) return [];` 
😬

server/repositories/cmsApi.js
andrewrlee: Pushed error handling of missing prison up to controller to fail hard as its more observable

server/routes/topics.js
andrewrlee: Fixed this in the middleware

server/repositories/cmsApi.js
andrewrlee: no longer relevant as edge case handling moved

server/content/establishmentData.json
```diff
@@ -20,9 +19,8 @@
   },
   "793": {
     "name": "wayland",
-    "formattedName": "Wayland",
-    "prefix": "HMP",
-    "uuId": "b73767ea-2cbb-4ad5-ba22-09379cc07241",
+    "displayName": "HMP Wayland",
+    "searchName": "Wayland",
```
andrewrlee, line 23: tidying to make these obvious and removing extra logic around concatenating prefix and name

server/middleware/configureEstablishment.js
```diff
@@ -19,18 +18,22 @@ const configureEstablishment = () => (req, res, next) => {
...
 
+    const establishmentId = getEstablishmentId(establishmentName);
+
     req.session.id = uuid();
-    req.session.establishmentName = establishmentName;
-    req.session.establishmentId = getEstablishmentId(establishmentName);
-    req.session.establishmentPersonalisationEnabled = getEstablishmentPersonalisation(
-      req.session.establishmentId,
-    );
+    if (typeof establishmentId === 'undefined') {
```
andrewrlee, line 24: we now only set establishment details if we can link to a real establishment.
Previously we would set the establishment name to `localhost:3000` as the regex is a little bit dodgy

server/middleware/configureEstablishment.js
```diff
@@ -19,18 +18,22 @@ const configureEstablishment = () => (req, res, next) => {
...
+      req.session.establishmentName = establishmentName;
+      req.session.establishmentPersonalisationEnabled = getEstablishmentPersonalisation(
+        establishmentId,
+      );
+    }
   }
 
   res.locals.feedbackId = uuid();
-  res.locals.establishmentDisplayName = `${getEstablishmentPrefix(
+  res.locals.establishmentDisplayName = `${getEstablishmentDisplayName(
```
andrewrlee, line 34: pushed this logic out of code and into config

server/repositories/cmsApi.js
```diff
@@ -1,37 +1,77 @@
...
+    'name',
+    'description',
+    'field_legacy_landing_page',
+  ])
+  .addFilter(
+    'vid.meta.drupal_internal__target_id',
+    ['moj_categories', 'tags'],
+    'IN',
+  )
+  .addSort('name')
```
andrewrlee, line 31: now sorting in drupal

server/routes/topics.js
```diff
@@ -1,41 +1,32 @@
...
     try {
-      const userName = req.user && req.user.getFullName();
-      const establishmentId = path(['session', 'establishmentId'], req);
-      const topics = await topicsService.getTopics(establishmentId);
+      const userName = req.user?.getFullName();
+      const { establishmentName } = req.session;
+
+      if (!establishmentName) {
+        throw new Error('Could not determine establishment!');
+      }
```
andrewrlee, line 15: rather than silently fail decided to hard fail as this should never happen and erroring is more observable

server/middleware/__tests__/configureEstablishment.spec.js
```diff
@@ -0,0 +1,41 @@
...
+
+  it('should use the set the session from the config', () => {
+    const req = {
+      session: {
+        establishmentName: 'wayland',
+        establishmentId: 793,
+      },
+    };
+
+    const res = { locals: {} };
```
jonhindle, line 18: Pop that `const res = { locals: {} };` in the beforeEach

server/middleware/__tests__/configureEstablishment.spec.js
```diff
@@ -0,0 +1,41 @@
...
+          'cookhamwood-prisoner-content-hub-development.apps.live-1.cloud-platform.service.justice.gov.uk/',
+      },
+      session: {},
+    };
+
+    const res = { locals: {} };
+
+    configureEstablishment()(req, res, next);
+
+    expect(res.locals.establishmentDisplayName).toBe(`HMYOI Cookham Wood`);
```
jonhindle, line 39: Minor linting question `'` or '`'

server/middleware/__tests__/configureEstablishment.spec.js
```diff
@@ -0,0 +1,41 @@
...
+    next.mockClear();
+  });
+
+  it('should use the set the session from the config', () => {
+    const req = {
+      session: {
+        establishmentName: 'wayland',
+        establishmentId: 793,
+      },
+    };
```
jonhindle, line 16: Is it worth running a precedent test, selection of the appropriate establishmentName from session over header? This could be incorporated into this test by providing erroneous header data?

jonhindle: meh, ignore this, I miss understood the code 👍

server/middleware/__tests__/configureEstablishment.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  beforeEach(() => {
+    next.mockClear();
+  });
+
+  it('should use the set the session from the config', () => {
+    const req = {
+      session: {
+        establishmentName: 'wayland',
+        establishmentId: 793,
+      },
```
jonhindle, line 15: I _think _ this test might be flawed 🤔 The establishment name from the header will return `wayland` and generate the id from that. The name and id being passed through are coming from that default and _not_ the session. I think. 🤯

server/middleware/configureEstablishment.js
```diff
@@ -15,22 +13,27 @@ const configureEstablishment = () => (req, res, next) => {
       req.session.establishmentPersonalisationEnabled === undefined)
   ) {
     const replaceUrl = /-prisoner-content-hub.*$/g;
-    const establishmentName = pathOr('wayland', ['headers', 'host'], req)
+
+    const establishmentName = (req.headers?.host || 'wayland')
```
jonhindle, line 17: Is a default required for this? Seemingly it masks a genuine issue?

server/repositories/__tests__/cmsApi.spec.js
```diff
@@ -1,11 +1,11 @@
...
-  prisons,
-  landingPage,
-} = require('../../../test/resources/primaryMenu');
+  jsonApiResponse,
+  category,
+  tag,
+} = require('../../../test/resources/topics');
 const { JsonApiClient } = require('../../clients/jsonApiClient');
-const { CmsApi } = require('../cmsApi');
+const { CmsApi, topicsQuery } = require('../cmsApi');
```
jonhindle, line 8: Is it worth actually putting a simple test in for `topicsQuery`? I appreciate its essentially a static constant, but wanted to ascertain if there was any value in highlighting any  fluctuations in the consistency of its value?

server/routes/topics.js
```diff
@@ -1,41 +1,32 @@
...
-const { path } = require('ramda');
 const Sentry = require('@sentry/node');
 const { logger } = require('../utils/logger');
 
-const fixUrls = element => {
-  const { id, description, href, linkText } = element;
-  switch (element.href) {
-    // TODO: Re-enable when Incentives personalization feature released
-    // case '/content/4204':
-    //   return { id, description, href: '/incentives', linkText };
```
jonhindle, line 11: Can you shed any light on this @davidthomas22?
It does look hack-y.

server/services/__tests__/topicsService.spec.js
```diff
@@ -18,70 +17,33 @@ describe('Topics Service', () => {
...
-
-    it('Sources to have been called correctly', async () => {
-      hubMenuRepo.tagsMenu.mockResolvedValue([]);
-      cmsApi.primaryMenu.mockResolvedValue([]);
-
-      await topicService.getTopics(1);
-
-      expect(hubMenuRepo.tagsMenu).toHaveBeenCalledWith(1);
-      expect(cmsApi.primaryMenu).toHaveBeenCalledWith(1);
+      expect(cmsApi.getTopics).toHaveBeenCalledWith('berwyn');
```
jonhindle, line 40: Finicky strikes again. `const ESTABLISHMENT = 'berwyn'` maybe? It crops up a few times 😬

server/services/search.js
```diff
@@ -2,10 +2,10 @@ const utils = require('../utils');
 
 const createSearchService = ({
   searchRepository,
-  getEstablishmentFormattedName = utils.getEstablishmentFormattedName,
+  getEstablishmentSearchName = utils.getEstablishmentSearchName,
 }) => {
   function find({ query, limit, from, establishmentId }) {
-    const prison = getEstablishmentFormattedName(establishmentId);
+    const prison = getEstablishmentSearchName(establishmentId);
```
jonhindle, line 8: Do we set `prison` first to highlight errors? Otherwise, it should probably go below the `query === ''` check following the whole _return early_ ideology.

server/services/search.js
```diff
@@ -15,7 +15,7 @@ const createSearchService = ({
   }
 
   function typeAhead({ query, limit, establishmentId }) {
-    const prison = getEstablishmentFormattedName(establishmentId);
+    const prison = getEstablishmentSearchName(establishmentId);
```
jonhindle, line 18: Same query as above...
Do we set `prison` first to highlight errors? Otherwise, it should probably go below the `query === ''` check following the whole return early ideology.

server/middleware/__tests__/configureEstablishment.spec.js
andrewrlee: I almost did this :-D

server/middleware/__tests__/configureEstablishment.spec.js
andrewrlee: no strong opinion - I would usually `'` for this so will change

server/middleware/__tests__/configureEstablishment.spec.js
andrewrlee: there's no establishment in the header here? 

I think you're right though it is wrong - the existing test just checked that it used the default prison:
`req.headers?.host || 'wayland'`

I'll split into 2 separate tests:
* one that uses the default 
* one that uses the session

(think we should ditch the defaulting behaviour everywhere though as it most likely going to silently show the wrong thing)

This was only existing test (the file didn't have `.spec.` in it so I've fixed it.

server/middleware/configureEstablishment.js
andrewrlee: ah.. just said that above 👍 

We'll leave and fix in a follow up - I'd feel much more comfortable stripping out all the defaulting to wayland

server/repositories/__tests__/cmsApi.spec.js
andrewrlee: it is implicitly used through nock... at one point I hardcoded the url that it generated in this test.. could potentially put that back in

server/services/search.js
andrewrlee: meh we don't error on it so no. 

I agree in principal but this is big enough already

server/services/search.js
andrewrlee: same as above.. totally agree but think we can leave for now?

server/services/__tests__/topicsService.spec.js
andrewrlee: meh.. little bit resistant to deduplication through use of constants in tests. Tend to like to see realistic examples in test data and use of constants can obscure that.

Not the strongest argument granted :-)

server/middleware/__tests__/configureEstablishment.spec.js
jonhindle: Yeah... I didn't expand the collapsed if statement 😬 In the right ballpark though...

jonhindle (dismissed):
Looks fine 👍

server/routes/topics.js
davidthomas22: I think this is related to old designs and is no longer is required

jonhindle (approved):
fine

PR #260, opened 2021-05-17
bug/fix-search: Changing search type

andrewrlee:
### Context

https://trello.com/c/AHyLgirl

We were previously doing a term search on a text value.
This has [known limitations](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html#avoid-term-query-text-fields)

 This would only work for single word prisons. The text value would be processed during index to split by individual word and lower case whereas the terms query would pass the exact phrase.

So when we searched for 'Cookham wood' it would fail to match either the tokens 'cookham' or 'wood'.
Luckily according to the index I could see that a separate  prison_name.keyword was also indexed which indexes the whole thing.

This change moves to using the formatted name and the keyword field as these currently always match.
We should see if drupal would allow us to index establishment id as going forward a keyword index on establishment id would be more robust.

### Considerations

Worth giving search a bit of a tire kicking generally across each prison

### DIFF
--- a/server/routes/__tests__/health.spec.js
+++ b/server/routes/__tests__/health.spec.js
@@ -1,14 +1,14 @@
 const request = require('supertest');
-const express = require('express')
+const express = require('express');
 
 const { createHealthRouter } = require('../health');
 
 describe('GET healthchecks', () => {
   let config;
   let app;
-  const BUILD_NUMBER = 'foo-number'
-  const GIT_REF = 'foo-ref'
-  const GIT_DATE = 'foo-date'
+  const BUILD_NUMBER = 'foo-number';
+  const GIT_REF = 'foo-ref';
+  const GIT_DATE = 'foo-date';
   beforeEach(() => {
     config = {
       buildInfo: {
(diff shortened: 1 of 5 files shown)


server/routes/__tests__/health.spec.js
```diff
@@ -1,14 +1,14 @@
 const request = require('supertest');
-const express = require('express')
+const express = require('express');
```
andrewrlee, line 2: fixing file getting past linter

server/routes/health.js
```diff
@@ -1,6 +1,6 @@
 const express = require('express');
 
-const createHealthRouter = ({ buildInfo:build }) => {
+const createHealthRouter = ({ buildInfo: build }) => {
```
andrewrlee, line 3: fixing file getting past linter

leonk (approved):
LGTM

davidthomas22 (approved):
Nice work, tested several searches on dev including misspelt words (Exercise, Music and, Guide, guide, bob and, christain, govenor) all returning expected values

PR #263, opened 2021-05-21
Removing jest config

andrewrlee:
### Intent

Remove need to set up project specific IDE settings to run tests and simplify consuming of env vars.  

### Considerations

Need to check retrieval of personalisation in dev - as some refactoring of urls


jonhindle (approved):
Everything is fine.

PR #283, opened 2021-06-09
Fixing staging deployment

andrewrlee:
In staging and prod there would be no qualifier so the helm release would be named `prisoner-content-hub-frontend-` which is invalid

This change only adds the `-` separator between release name and qualifier if qualifier is present

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,9 @@ commands:
           command: |
             VERSION_TO_DEPLOY=$(cat /tmp/build-info/version-to-deploy.txt)
             sed -i "s/appVersion:.*/appVersion: \"${VERSION_TO_DEPLOY}\"/g" "./helm_deploy/prisoner-content-hub-frontend/Chart.yaml"
-            helm upgrade << parameters.releaseName >>-<< parameters.qualifier >> ./helm_deploy/prisoner-content-hub-frontend \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -63,7 +63,9 @@ commands:
           command: |
             VERSION_TO_DEPLOY=$(cat /tmp/build-info/version-to-deploy.txt)
             sed -i "s/appVersion:.*/appVersion: \"${VERSION_TO_DEPLOY}\"/g" "./helm_deploy/prisoner-content-hub-frontend/Chart.yaml"
-            helm upgrade << parameters.releaseName >>-<< parameters.qualifier >> ./helm_deploy/prisoner-content-hub-frontend \
+            QUALIFIER=<< parameters.qualifier >>
+            RELEASE_NAME=<< parameters.releaseName >>${QUALIFIER:+-$QUALIFIER}
+            helm upgrade $RELEASE_NAME ./helm_deploy/prisoner-content-hub-frontend \
```
andrewrlee, line 68: weird syntax is bash string function:: 
https://stackoverflow.com/questions/40771781/how-to-append-a-string-in-bash-only-when-the-variable-is-defined-and-not-null

jonhindle (approved):
lgtm

PR #293, opened 2021-06-17
Moving to multi-part build and node buster image to reduce image size

andrewrlee:
### Context

https://trello.com/c/0C1qINgT

### Intent

Speed up build and reduce docker image by moving to a multi part build 

Docker stores the state of each layer in the final image. By moving to a multi stage build you can effectively remove the size of earlier stages from the total image size.

Also moving to a more standard (and slightly later) node image.

This knocks at least a minute (maybe 2?) off the build and reduces the image size from just ~1GB to ~275MB

### Considerations

This is a fundamental change to how the application is built so could do with rigorous testing

### Checklist

- [ ] Tested in Development


Dockerfile
```diff
@@ -1,42 +1,42 @@
...
-COPY package*.json ./
-RUN CYPRESS_INSTALL_BINARY=0  npm ci --no-audit
+RUN apt-get update && \
+    apt-get upgrade -y
+
+WORKDIR /app
+
+COPY . .
+
+RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit --production && npm run build
```
mattops, line 15: I dont think there is any point in chaining these commands because the builder stage will be discarded anyway.

mattops (approved):
LGTM

PR #294, opened 2021-06-21
Removing unused components

andrewrlee:
### Context

Removing components that are no longer being used

### Considerations

These were personalisation components that were not removed when we removed the content/personalisation hybrid pages

### Checklist

- [ ] Tested in Development


jonhindle (approved):
Good pruning that man!

PR #295, opened 2021-06-22
feature/display-number-of-visits-remaining

jonhindle:
### Context

https://trello.com/c/dP0bbgEO

Navigate to the profile page and click sign in.

### Intent

Additional API call to get the visits balance and display tweaks.

Screenshot 2021-06-22 at 16 46 22

### Considerations

Errors from either visits API call should result in the error message displaying

N/A

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/routes/profile.js
+++ b/server/routes/profile.js
@@ -4,13 +4,19 @@ const createProfileRouter = ({ offenderService }) => {
   const router = express.Router();
 
   const getPersonalisation = async user => {
-    const [eventsSummary, incentivesSummary, visitsSummary, moneySummary] =
-      await Promise.all([
-        offenderService.getEventsForToday(user),
-        offenderService.getIncentivesSummaryFor(user),
-        offenderService.getVisitsFor(user),
-        offenderService.getBalancesFor(user),
-      ]);
+    const [
+      eventsSummary,
+      incentivesSummary,
+      visitsSummary,
+      visitsBalances,
+      moneySummary,
+    ] = await Promise.all([
+      offenderService.getEventsForToday(user),
+      offenderService.getIncentivesSummaryFor(user),
+      offenderService.getVisitsFor(user),
+      offenderService.getVisitBalances(user),
+      offenderService.getBalancesFor(user),
+    ]);
 
     const signedInUser = user.getFullName();
 
@@ -37,6 +43,9 @@ const createProfileRouter = ({ offenderService }) => {
       error: visitsError,
     } = visitsSummary;
 
+    const { visitsRemaining = 0, error: visitsRemainingError = false } =
+      visitsBalances;
+
     const { spends, privateAccount, savings, error: moneyError } = moneySummary;
 
     return {
@@ -56,7 +65,7 @@ const createProfileRouter = ({ offenderService }) => {
         link: '/content/4204',
       },
       visitsSummary: {
-        error: visitsError,
+        error: visitsError || visitsRemainingError,
         hasNextVisit,
         nextVisit,
         visitType,
@@ -64,6 +73,7 @@ const createProfileRouter = ({ offenderService }) => {
         startTime,
         endTime,
         link: '/content/4203',
+        visitsRemaining,
       },
       moneySummary: {
         error: moneyError,
--- a/server/services/__tests__/offenderService.spec.js
+++ b/server/services/__tests__/offenderService.spec.js
@@ -340,7 +340,7 @@ describe('Offender Service', () => {
 
     it('should return an error when no booking Id is passed', async () => {
       const repository = {
-        getEventsFor: jest.fn().mockReturnValue([]),
+        getEventsFor: jest.fn().mockReturnValue({ error: true }),
       };
 
       const service = createPrisonApiOffenderService(repository);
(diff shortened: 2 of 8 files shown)


server/routes/profile.js
```diff
@@ -56,14 +68,15 @@ const createProfileRouter = ({ offenderService }) => {
...
-        error: visitsError,
+        error: visitsError || visitsRemainingError,
         hasNextVisit,
         nextVisit,
         visitType,
         visitorName,
         startTime,
         endTime,
         link: '/content/4203',
+        visitsRemaining: remainingPvo + remainingVo,
```
andrewrlee, line 79: meh.. could centralize this logic by potentially pushing it down to the service layer.  

`getVisitsRemaining` rather than `getVisitsBalance`

andrewrlee (approved):
LGTM - haven't tested on dev yet

server/routes/profile.js
```diff
@@ -37,6 +43,12 @@ const createProfileRouter = ({ offenderService }) => {
       error: visitsError,
     } = visitsSummary;
 
+    const {
+      remainingPvo = 0,
+      remainingVo = 0,
+      error: visitsRemainingError = false,
```
andrewrlee, line 49: i guess if you did push down that logic you could possibly get away without the defaulting here

service layer returning  `{ error: true } | { count: number }`   🤷

server/services/offender/index.js
```diff
@@ -116,6 +116,25 @@ const createOffenderService = (
     }
   }
 
+  async function getVisitBalances({ prisonerId }) {
+    try {
+      logger.info(`OffenderService (getVisitBalances) - User: ${prisonerId}`);
+
+      if (!prisonerId) {
+        throw new Error('No prisonerId passed');
+      }
```
andrewrlee, line 125: I know this is based on existing patterns but not a fan of these preconditions , especially considering it gets caught by the surrounding try.

Not sure what you think, I guess I would prefer we didn't even call the service layer if we don't have a prisoner ID - I imagine it never actually happens due to the way we only add personalisation if we have a logged in user  so could just get rid?

Also not sure why we don't just log the stack trace in the catch.

Could we ditch the debug of stack trace if we capture using sentry?

server/routes/profile.js
jonhindle: I originally did and changed it to return all the values. I guess it boils down to where you want the business logic to lie. I figured this way, the decision about which data to use lies within _profile.js_.

server/routes/profile.js
jonhindle: I don't think either defaults are required here. They're there for belt and braces I guess 🤷

server/services/offender/index.js
jonhindle: Good call; this check is pointless. Because I've destructed, the error would already have blown this function up.

davidthomas22 (approved):
Looks good to me

server/services/__tests__/offenderService.spec.js
```diff
@@ -340,7 +340,7 @@ describe('Offender Service', () => {
 
     it('should return an error when no booking Id is passed', async () => {
       const repository = {
-        getEventsFor: jest.fn().mockReturnValue([]),
+        getEventsFor: jest.fn().mockReturnValue({ error: true }),
```
andrewrlee, line 343: very meh... guess this should be resolved rather than return as this will return a promise

server/services/offender/index.js
```diff
@@ -116,21 +96,34 @@ const createOffenderService = (
     }
   }
 
-  async function getImportantDatesFor(user) {
+  async function getVisitBalances({ prisonerId }) {
```
andrewrlee, line 99: do you have a service layer test for this?

server/services/__tests__/offenderService.spec.js
jonhindle: So Async automatically wraps any response in a Promise I think. There are 21 other instances of _ mockReturnValue_ in this test block where the function that is mocked is _async_. From what I can see all the _real_ functions end up being auto wrapped, so I guess that was the reasoning in the existing tests? I guess returning a Promise would ensure that we were unwrapping with an _await_ or _then_ though? 
I'm easy either way 🤷

andrewrlee: yes basically - its the kind of thing that typescript would have detected.

I have seen (done) bugs where people (me) have forgotten to await promises and tests have passed - but here you need to remember to mockResolve so its kind of a double accounting

server/services/__tests__/offenderService.spec.js
```diff
@@ -116,6 +116,28 @@ describe('Offender Service', () => {
...
+      const service = createPrisonApiOffenderService(repository, {
+        NextVisit: mockAdapter,
+      });
+
+      const data = await service.getVisitsRemaining(TEST_USER);
+
+      expect(repository.getVisitBalances).toHaveBeenCalledWith(
+        TEST_PRISONER_ID,
+      );
+      expect(mockAdapter.from).toHaveBeenCalledWith(RAW_RESPONSE);
```
andrewrlee, line 136: how does this work? You're not using NextVisit are you? 

wonder if there's some kind of test set up issues

andrewrlee (approved):
LGTM

davidthomas22 (approved):
LGTM

PR #296, opened 2021-06-22
Enabling mock auth in dev

andrewrlee:
### Context

https://trello.com/c/gt21UBzk

### Intent

We have done the work to create an environment per PR but unfortunately we cannot get production style auth working with these as it doesn't support dynamic URLs

We've agreed to use mock auth in dev instead and this change enables that. 

This will mean that we can actually  provision different users on different PRs or potentially have somekind of  user switcher.

### Considerations

Going forward we will only be able to test auth in staging
Need to ensure that mock auth is DISABLED in staging and prod during the release.

### Checklist

- [ ] Tested in Development


jonhindle (approved):
LGTM

davidthomas22 (commented):
LGTM

PR #334, opened 2021-08-06
Supporting basic pages

andrewrlee:
### Context

https://trello.com/c/fh1sP4L6

### Intent

Moving to use JSON API for basic pages 

### Considerations

This moves leaves other content remaining on the old API

### Checklist

- [ ] Tested in Development


server/repositories/cmsApi.js
```diff
@@ -15,11 +15,29 @@ class CmsApi {
     this.jsonApiClient = jsonApiClient;
   }
 
+  async lookupContent(contentId) {
+    // this will need to use config to get host name
+    const response = await this.jsonApiClient.get(
+      `https://cms-prisoner-content-hub-development.apps.live-1.cloud-platform.service.justice.gov.uk/router/translate-path?path=content/${contentId}`,
```
andrewrlee, line 21: this will eventually include establishment name when Leon makes changes

server/config.js
```diff
@@ -40,12 +40,14 @@ module.exports = {
   api: {
     hubEndpoint,
     hubContent: `${hubEndpoint}/v1/api/content`,
+    hubContentRouter: `${hubEndpoint}/router/translate-path`,
```
andrewrlee, line 43: would not worry about adding to this - these paths aren't configuration so shouldn't really be here and can go into the client. 

So would just use `hubEndpoint` and hardcode `/router/translate-path` in the client

server/repositories/cmsApi.js
```diff
@@ -15,10 +15,29 @@ class CmsApi {
...
+    };
+  }
+
   async get(query) {
-    const response = await this.jsonApiClient.get(query.path());
-    const items = dataFormatter.deserialize(response);
-    return items.map(item => query.transform(item));
+    const request = query.path
+      ? this.jsonApiClient.getRelative(query.path())
+      : this.jsonApiClient.getUrl(query.url());
```
andrewrlee, line 34: allowing queries to either specify absolute or relative urls.

 `/content/:id` items will be looked up via the router url above and that will provide information about the type of item and also it's absolute location.

server/repositories/cmsApi.js
```diff
@@ -15,10 +15,29 @@ class CmsApi {
...
+    const response = await request;
+    const deserializedResponse = dataFormatter.deserialize(response);
+    if (Array.isArray(deserializedResponse)) {
```
andrewrlee, line 37: when retrieving a specific item referenced by the router, it will point to a single item rather than a list - this will handle the difference transparently.

We could have added a different method to `getSingle` but we didn't want to add confusion to the calling code - where the calling code would have to know which method to call based on some intrinsic aspect of the query

server/services/cms.js
```diff
@@ -8,6 +9,22 @@ class CmsService {
...
+      establishmentName,
+      contentId,
+    );
+
+    switch (type) {
+      case 'node--page':
+        return this.#cmsApi.get(new BasicPageQuery(location));
+      /// ...other types go here
+      default:
+        // log unsupported type
```
andrewrlee, line 23: when we move all content items into here, we can remove this null check and throw on unexpected types

server/views/pages/flat-content.html
```diff
@@ -43,7 +43,7 @@
           <h1 id="title" class="govuk-heading-xl govuk-!-margin-bottom-6 govuk-!-margin-top-3">{{data.title}}</h1>
           <p id="stand-first" class="govuk-body-l">{{data.standFirst}}</p>
           <div id="body" class="gov-uk-dynamic-content">
-            {{data.description.sanitized | safe }}
+            {{data.description | safe }}
```
andrewrlee, line 46: simplfied this by concealing complexity of underlying item response from drupal

server/routes/content.js
```diff
@@ -23,9 +23,14 @@ const createContentRouter = ({ hubContentService, analyticsService }) => {
...
     const establishmentId = path(['session', 'establishmentId'], req);
     const userAgent = path(['headers', 'user-agent'], req);
+    const { establishmentName } = req.session;
 
     try {
-      const data = await hubContentService.contentFor(id, establishmentId);
+      const data = await hubContentService.contentFor(
+        id,
+        establishmentId,
+        establishmentName,
```
andrewrlee, line 32: we'll eventually swap this out and just call the cms service and only pass establishment name

server/config.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
     series: `${hubEndpoint}/v1/api/content/series`,
     tags: `${hubEndpoint}/v1/api/vocabulary/tags`,
   },
+
```
jonhindle, line 49: In keeping with the rest of this file, I would remove this addition.

davidthomas22 (approved):
LGTM

PR #359, opened 2021-09-13
Allowing berwyn to access all of the profile apart from visits section

andrewrlee:
### Context

Adding the profile to berwyn but without the visits row

### Intent

https://trello.com/c/TRCiCgJ2

Now no longer representing personalisation as a distinct feature - we want this to be present across all prisons. 
 
As part of this we've also decided to remove the coronavirus section from popular topics and rename music and talk to NPR - to match the renaming of content done the other week

### Considerations

We could have represented showing visits as a separate feature to enable of disable at an establishment. Decided not to do this as we don't have any intention of supporting this in the future. Berwyn is just the odd one out in this case. 

### Checklist

- [ ] Tested in Development


psoleckimoj (approved):
LGTM!

server/middleware/configureEstablishment.js
```diff
@@ -24,12 +15,11 @@ const configureEstablishment = () => (req, res, next) => {
...
-      req.session.establishmentPersonalisationEnabled =
-        getEstablishmentPersonalisation(establishmentId);
     }
   }
 
   res.locals.feedbackId = uuid();
+  res.locals.establishmentName = req.session.establishmentName;
   res.locals.establishmentDisplayName = `${getEstablishmentDisplayName(
     req.session.establishmentId,
   )}`;
```
jonhindle, line 25: I appreciate this is out of scope but why is this a string variable? A glorified toString()?

server/middleware/configureEstablishment.js
```diff
@@ -1,17 +1,8 @@
...
-  if (
-    req.session &&
-    (!req.session.id ||
-      !req.session.establishmentId ||
-      req.session.establishmentPersonalisationEnabled === undefined)
-  ) {
+  if (req.session && (!req.session.id || !req.session.establishmentId)) {
     const replaceUrl = /-prisoner-content-hub.*$/g;
 
     const establishmentName = (req.headers?.host || 'wayland')
```
jonhindle, line 8: I thought we decided to ditch the auto `wayland` behaviour?

jonhindle (approved):
I don't really like the `if berwyn` elements of the code; it doesn't allow for other establishments opting for a diminished functionality. Although I think I object to us catering 
 for this level of selective degradation more 🤷 
lgtm 👍

server/middleware/configureEstablishment.js
andrewrlee: dunno

server/middleware/configureEstablishment.js
andrewrlee: yes, this is becoming slowly less relevant you can't access the homepage without an establishment now and haven't been able to for a couple of months or so. 

This will be safe to remove entirely going forward

andrewrlee:
> it doesn't allow for other establishments opting for a diminished functionality, Although I think I object to us catering
for this level of selective degradation more

Yes, this is precisely how I feel about it

PR #360, opened 2021-09-13
Moving to run integration tests using docker-compose

andrewrlee:
### Context

This moves integration tests over to be run with docker-compose. This allows us to use docker networking to have tests that run in Circle CI against a specific prison url, e.g: http://cookhamwood.prisoner-content-hub.local:3000

This is important as we currently derive the establishment from the hostname.

Also this should help when we add integration tests with drupal.  

### Intent

https://trello.com/c/txfklMBK

### Considerations

One of the benefits of this approach is that we can move all the cypress test dependencies out of the main project and into a separate one. 

I've currently moved most of the existing tests that don't run (and can't run) into another folder so they won't be picked up.

Maintained a separate command to run tests with the UI. 

This does currently have the disadvantages:
* rebuilds the docker image again (so a minute or so slower!)
* Correctly fails build when tests fail but doesn't publish test results

### Checklist

- [ ] Tested in Development


.circleci/config.yml
```diff
@@ -167,29 +167,17 @@ jobs:
...
-          name: Run the node app.
-          command: npm run start:e2e
-          background: true
-      - run:
-          name: Get wiremock
-          command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-standalone/2.27.2/wiremock-standalone-2.27.2.jar
-      - run:
-          name: Run wiremock
-          command: java -jar wiremock.jar --port 9091
-          background: true
```
andrewrlee, line 180: now running wiremock through docker compose and don't need to worry about running the app

andrewrlee: currently rebuilds docker image rather than using the newly built one - would be nice to fix this

.circleci/config.yml
```diff
@@ -167,29 +167,17 @@ jobs:
...
-          name: Wait for node app to start
-          command: sleep 5
+      - setup_remote_docker:
+          docker_layer_caching: true
       - run:
           name: e2e tests
           command: npm run test:e2e:ci
       - store_test_results:
-          path: test-results
+          path: e2e/test-results
```
andrewrlee, line 176: this currently isn't working as the artifacts are inside the cypress docker image - still need to copy them out. 
It correctly reports when the tests fail though.

Dockerfile
```diff
@@ -12,7 +12,7 @@ WORKDIR /app
 
 COPY . .
 
-RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit --production && \
+RUN npm ci --no-audit --production && \
```
andrewrlee, line 15: don't have to worry about this as cypress no longer present in prod dependencies

e2e/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM cypress/base:10
+WORKDIR /app
+
+COPY package.json .
+COPY package-lock.json .
+COPY reporter-config.json .
+COPY cypress.json .
+ADD cypress cypress
```
andrewrlee, line 8: need to manually make these resources on as docker-compose in circle doesn't support mounted volumes.

e2e/cypress/mockApis/wiremock.js
```diff
@@ -1,6 +1,9 @@
 const superagent = require('superagent');
 
-const url = 'http://localhost:9091/__admin';
+const host = process.env.WIREMOCK_BASE_URL || 'http://localhost:9091';
```
andrewrlee, line 3: allow overriding wiremock url as it will differ between running in docker compose (which uses docker networking) and outside it.

e2e/cypress/support/step_definitions/approved-features.js
```diff
@@ -1,5 +1,4 @@
 import { Before, Then, When, And } from 'cypress-cucumber-preprocessor/steps';
-import { expectCt } from 'helmet';
```
andrewrlee, line 2: not used!

e2e/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+          - wayland.prisoner-content-hub.local
+          - berwyn.prisoner-content-hub.local
+          - cookhamwood.prisoner-content-hub.local
+    depends_on:
+      - wiremock
+    restart: always
+    ports:
+      - "3000:3000"
+    env_file: 
+      - "./test-ci.env"
```
andrewrlee, line 18: now two separate test env files - one for running locally and one for inside circle

package.json
```diff
@@ -13,10 +13,8 @@
     "lint": "eslint . --cache --max-warnings 0",
     "test": "jest",
     "clear_jest": "jest --clearCache",
-    "start:e2e": "export $(cat feature.env) && node server.js",
-    "test:e2e:run": "start-server-and-test start:e2e http://localhost:3000 test:e2e:ci",
-    "test:e2e:ci": "cypress run --spec cypress/features/timetable.feature,cypress/features/profile.feature,cypress/features/approved-visitors.feature",
-    "test:e2e:ui": "cypress open",
+    "test:e2e:ci": "npm run start-ci --prefix ./e2e",
+    "test:e2e:ui": "npm run start-ui --prefix ./e2e",
```
andrewrlee, line 17: encapsulating complexity of running int tests in sub project

jonhindle (approved):
lgtm 🤷‍♀️

PR #361, opened 2021-09-13
feature/switch-to-JSONAPI-for-series-pages

jonhindle:
### Context

https://trello.com/c/EyX2RFzY/112-switch-to-jsonapi-for-series-pages

N/A

### Intent

We no longer serve series 

no

### Considerations

original code left in for categories

no

### Checklist

- [ ] Tested in Development


server/repositories/cmsQueries/seriesHeaderPageQuery.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  transform(item) {
+    return {
+      id: item?.drupalInternal_Tid,
+      contentType: 'series',
+      name: item?.name,
+      description: item?.description?.processed,
+      image: {
+        url: item?.fieldFeaturedImage?.imageStyleUri[0]?.tile_large,
```
andrewrlee, line 29: I'm not sure the order of the images is safe to assume here.

andrewrlee: I wonder if Leon could modify the structure so it's easier to consume

server/repositories/cmsQueries/seriesPageQuery.js
```diff
@@ -0,0 +1,81 @@
...
+      contentUrl: `/content/${id}`,
+      image: {
+        url: item?.fieldMojThumbnailImage?.imageStyleUri[1].tile_small,
+        alt: item?.fieldMojThumbnailImage?.resourceIdObjMeta?.alt,
+      },
+    };
+  };
+
+  transform(deserializedResponse) {
+    return Object.assign(
```
andrewrlee, line 69: Meh.. I was doing a bit of reading around this.. It doesn't look like there is a good reason to use `Object.assign` for places where object spread would suffice. 

As well as spread being more succinct, I think it's generally good to prefer approaches that don't mutate existing data.

The only reason to use Object.assign seems to be if you want to actually call setters or preserve the prototype of the original object - would possibly only reserve it to be used in those cases.

server/routes/tags.js
```diff
@@ -29,7 +29,9 @@ const createTagRouter = ({ hubTagsService }) => {
 
       data.secondaryTags = data.id;
 
-      const pageType = data.contentType === 'tags' ? 'tags' : 'tagsSeries';
+      const pageType = ['tags', 'series'].includes(data.contentType)
+        ? 'tags'
+        : 'tagsCategories';
```
andrewrlee, line 34: I thought we only supported secondary tags and series on this route? Do we need `tagsCategories`?

server/services/cms.js
```diff
@@ -32,14 +38,25 @@ class CmsService {
...
+
+  async getTag(establishmentName, id) {
+    const lookupData = await this.#cmsApi.lookupTag(establishmentName, id);
     const { type, uuid, location } = lookupData;
     switch (type) {
       case 'taxonomy_term--tags':
         return this.getSecondaryTag(establishmentName, uuid, location);
       case 'taxonomy_term--series':
-        return null; // TODO
+        return this.getSeries(establishmentName, id, location);
```
andrewrlee, line 59: we could possibly throw from default now if we come across some type of content we wouldn't expect to serve?

server/repositories/cmsQueries/homePageQuery.js
```diff
@@ -49,10 +50,10 @@ class HomepageQuery {
 
   transformEach(item) {
     const [upperFeatured, lowerFeatured] = item.fieldMojFeaturedTileLarge.map(
-      this.#asTile('tile_large'),
+      this.#asTile(getLargeTile),
```
andrewrlee, line 53: meh.. maybe not worth holding this up but could possible get rid of the `uri` from the file fields as presumably no longer used?

server/repositories/cmsQueries/videoPageQuery.js
```diff
@@ -57,11 +58,6 @@ class VideoPageQuery {
     }));
 
   transform(item) {
-    const url =
-      item.fieldMojThumbnailImage?.imageStyleUri
-        ?.map(image => image?.tile_large)
-        .find(image => Boolean(image)) || item.fieldMojThumbnailImage?.uri?.url;
```
andrewrlee, line 63: same here regarding removing URI from fields.

server/repositories/cmsQueries/audioPageQuery.js
```diff
@@ -58,11 +59,6 @@ class AudioPageQuery {
     }));
 
   transform(item) {
-    const url =
```
andrewrlee, line 61: same here round uri

server/repositories/cmsQueries/seriesPageQuery.js
jonhindle: The use of spread or assign here is negligible. The first object gets created and torn down and then recreated in this scenario with a spread, theoretically it only gets created once with Object.assign. I'm easy either way though 🤷

server/routes/tags.js
jonhindle: In the interests of maintaining existing functionality, we should leave it. Raise a ticket to remove, that way any changes in functionality is coming from product.

server/services/cms.js
jonhindle: Agreed, but we still have to cater for categories atm.

server/repositories/cmsQueries/__tests__/homePageQuery.spec.js
```diff
@@ -22,7 +22,9 @@ describe('Homepage query', () => {
             title: 'Lower Abs workout',
             fieldMojDescription: { summary: 'Intense lower core workout' },
             fieldMojThumbnailImage: {
-              uri: { url: 'https://cloud-platform-c3b3.eu-west-2' },
+              imageStyleUri: [
+                { tile_small: 'https://cloud-platform-c3b3.eu-west-2' },
+              ],
```
andrewrlee, line 27: couple of tests we could remove heree:
```
it('should show small tile image when provided', async () => {
it('should show large tile image when provided', async () => {
```

now we know that they will always exist

andrewrlee (approved):
LGTM

PR #363, opened 2021-09-16
Exclude assets and ping

andrewrlee:
### Context
bodged this earlier

### Intent

Need || rather than && !


psoleckimoj (approved):
LGTM, sorry for not noticing!

PR #366, opened 2021-09-17
Preventing missing content from triggering alerts

andrewrlee:
### Context

https://trello.com/c/AGF41AVL

Try and access content that doesn't exist: `/content/123456` or `/tag/123456`
Or content that is not available for a specific prison: `/content/5925` in berwyn

### Intent

We now capture 404 errors when calling queries and convert them to 404 responses sent to the browser.
The router returns 403 responses when content exists but not at a specific prison - we are now handling these as well

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -26,6 +26,7 @@ const defaultConfig = require('./config');
 const defaultEstablishmentData = require('./content/establishmentData.json');
 const defaultAuthMiddleware = require('./auth/middleware');
 const routes = require('./routes');
+const { NotFound } = require('./repositories/apiError');
 
 const createApp = services => {
   const {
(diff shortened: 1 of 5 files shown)


server/app.js
```diff
@@ -260,17 +261,19 @@ const createApp = services => {
...
+    if (error instanceof NotFound) {
+      logger.warn(`Failed to find: ${error.message}`);
+      logger.debug(error.stack);
+      res.status(404);
+      return res.render('pages/404');
+    }
+
     logger.error(`Unhandled error - ${req.originalUrl} - ${error.message}`);
     logger.debug(error.stack);
-    res.status(error.status || 500);
```
andrewrlee, line 265: generally shouldn't propagate downstream http statuses to the browser - they should generally be interpreted as 500s unless actually handled

server/repositories/hubContent.js
```diff
@@ -18,7 +18,7 @@ const hubContentRepository = httpClient => {
     const endpoint = `${config.api.hubContent}/${id}`;
 
     if (!id) {
-      logger.error(`HubContentRepository (contentFor) - No ID passed`);
+      logger.info(`HubContentRepository (contentFor) - No ID passed`);
```
andrewrlee, line 21: reducing these to `info`. These currently spam the logs and only represent a linked entity not existing - e.g if there is no associated series with a piece of content. So not actually errors

jonhindle (approved):
lgtm

PR #372, opened 2021-09-24
Removing profile from berwyn

andrewrlee:
### Context

https://trello.com/c/RehSBZ6F

### Intent

Removing profile from Berwyn as users cannot authenticate.


jonhindle (approved):
lgtm

PR #393, opened 2021-11-10
Moving to use establishment name rather than search name

andrewrlee:
### Context

No, part of the wider work around addressing tech debt to help with move to a single url

### Intent

Starting to remove reliance on Establishment ID by moving the search over to use establishment in name

(I think establishment ID might be a better name that establishment Name to represent what we call "machine name" in drupal?, but think it would be good to remove all references to ID before moving over to use it )

### Considerations

Establishment Name should always be set in the session on the search page - either from the user or from the domain name 

### Checklist

- [ ] Tested in Development


jonhindle (approved):
lgtm

jonhindle (commented):
LGTM

PR #400, opened 2021-11-18
Removing old tag handling code

andrewrlee:
### Context

Removing legacy code for calling legacy bespoke drupal api

### Intent

Removing old code and now throwing an error if we try to handle unknown content types

### Checklist

- [ ] Tested in Development


jonhindle (approved):
LGTM

PR #412, opened 2021-12-30
Send correct hostname to google analytics

andrewrlee:
Fix case when single url is used

### Context

https://trello.com/c/j68vCqlK

When we are using the single prison agnostic url we still want to be able to track per establishment usage via google analytics.

### Intent

This builds up the url based on the session derived establishment id and the single host url. 

### Considerations

URL pattern differs between dev/staging and prod: `-` vs `.` separated.

This doesn't take that into account so will start to report different urls in dev and staging.
I don't think this matters but need to check

### Checklist

- [ ] Tested in Development

### DIFF
--- a/helm_deploy/prisoner-content-hub-frontend/templates/_envs.tpl
+++ b/helm_deploy/prisoner-content-hub-frontend/templates/_envs.tpl
@@ -93,5 +93,9 @@ env:
         secretKeyRef:
           name: {{ include "prisoner-content-hub-frontend.fullname" . }}
           key: azureAdClientSecret
-
+      
+    - name: SINGLE_HOST_NAME
+    {{- with .Values.ingress.host }}
+      value: {{ tpl .pattern (dict "qualifier" $.Values.ingress.qualifier "Template" $.Template) }}
+    {{- end }}
 {{- end -}}
--- a/server/config.js
+++ b/server/config.js
@@ -1,26 +1,19 @@
-const { getEnv, isProduction } = require('../utils/index');
+const { getEnv, getRequiredEnv, isProduction } = require('../utils/index');
 
-const hubEndpoint = getEnv('HUB_API_ENDPOINT', { requireInProduction: true });
-const hmppsAuthBaseUrl = getEnv('HMPPS_AUTH_BASE_URL', 'https://api.nomis', {
-  requireInProduction: true,
-});
-const elasticsearchEndpoint = getEnv(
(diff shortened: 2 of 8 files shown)


helm_deploy/prisoner-content-hub-frontend/templates/_envs.tpl
```diff
@@ -93,5 +93,9 @@ env:
         secretKeyRef:
           name: {{ include "prisoner-content-hub-frontend.fullname" . }}
           key: azureAdClientSecret
-
+      
+    - name: SINGLE_HOST_NAME
```
andrewrlee, line 97: make single hostname available as env var

server/config.js
```diff
@@ -30,11 +23,10 @@ module.exports = {
     gitDate: getEnv('GIT_DATE', '2020-06-21 12:12:12'),
   },
   cookieSecret: getEnv('COOKIE_SECRET', 'keyboard cat'),
+  singleHostName: getRequiredEnv('SINGLE_HOST_NAME', 'localhost'),
   auth: {
-    clientId: getEnv('AZURE_AD_CLIENT_ID', { requireInProduction: true }),
-    clientSecret: getEnv('AZURE_AD_CLIENT_SECRET', {
-      requireInProduction: true,
-    }),
```
andrewrlee, line 37: these were broken as they required a default var so used the ` { requireIn...` as the default value.
Extracted out a `getRequiredEnv ` instead and provided noddy defaults

server/routes/analytics.js
```diff
@@ -20,10 +21,11 @@ const createAnalyticsRouter = ({ analyticsService }) => {
   });
 
   router.post('/page', (req, res) => {
-    const sessionId = path(['session', 'id'], req);
+    const sessionId = req.session.id;
+    const hostname = `${req.session.establishmentName}.${config.singleHostName}`;
```
andrewrlee, line 25: now build up host name from derived establishment name and single hostname

test/test-helpers.js
```diff
@@ -4,6 +4,8 @@ const nunjucksSetup = require('../server/utils/nunjucksSetup');
 function setupBasicApp(config = {}) {
   const app = express();
   app.set('view engine', 'html');
+  app.use(express.json());
+  app.use(express.urlencoded({ extended: true }));
```
andrewrlee, line 8: tests required parsing of json

jonhindle (approved):
LGTM

PR #417, opened 2022-01-11
Log user id of those making feedback statements

andrewrlee:
### Context

https://trello.com/c/4ERfhCSK

### Intent

Link users to feedback if currently logged in

### Considerations

Will wait for confirmation that this is desired. 

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/routes/__tests__/feedback.spec.js
+++ b/server/routes/__tests__/feedback.spec.js
@@ -1,0 +1,73 @@
+const request = require('supertest');
+const { createFeedbackRouter } = require('../feedback');
+const {
+  setupBasicApp,
+  consoleLogError,
+} = require('../../../test/test-helpers');
+
+describe('POST /feedback', () => {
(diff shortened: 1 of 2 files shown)


server/routes/__tests__/feedback.spec.js
```diff
@@ -0,0 +1,73 @@
...
+            categories: body.categories,
+            comment: body.comment,
+            contentType: body.contentType,
+            date: expect.anything(),
+            secondaryTags: body.secondaryTags,
+            sentiment: body.sentiment,
+            series: body.series,
+          });
+        });
+    });
```
jonhindle, line 71: Do we also need to add a test to confirm it logs, baring in mind this is how we intend to map feedback to accounts.

jonhindle (approved):
One query regarding testing that the ids are logged, otherwise lgtm

PR #459, opened 2022-03-29
Sanitising and limiting search input

andrewrlee:
### Context

https://trello.com/c/hYbFqugJ/

### Intent

The following changes:
1.) limit frontend and typeahead search to only allow queries of up to 50 characters.
2.) Backend, never execute queries of over 50 characters, just return empty list instead
3.) Remove any character in a query which doesn't match one of the following characters a-zA-Z0-9';,\-()!"

### Considerations

We noticed that the search: `Using Unilink: Shopping / canteen` currently throws an error on the backend and shows no results on the frontend. If we strip out the `/` then it works and shows the correct result.

(This also works if we strip out both the / and the : which is why I'm not allowing : either) (edited)

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/services/__tests__/search.spec.js
+++ b/server/services/__tests__/search.spec.js
@@ -37,6 +37,16 @@ describe('SearchService', () => {
       expect(results.length).toBe(0);
       expect(cmsApi.get).not.toHaveBeenCalled();
     });
+
+    it('returns empty when invalid', async () => {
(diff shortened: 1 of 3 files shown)


davidthomas22 (approved):
LGTM 

Nice spot on the sanitise search query

PR #475, opened 2022-05-05
Add correct links for content

andrewrlee:
This fixes the link to links

### Context

https://trello.com/c/EwL0hwwC

### Intent

Rather than trying to guess the content that the url will be present in the hub, instead use the path as provided from drupal.

### Considerations

Think the search index needs rebuilding on dev, so might be easier to test elsewhere.

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

PR #484, opened 2022-05-23
Feature/create show most recent content page

BarryGeeMOJ:
### Context

https://trello.com/c/KRCuHwwR

n/a

### Intent

new page and logic to display required data.

n/a

### Considerations

n/a

n/a
### Checklist

- [ ] Tested in Development


server/repositories/cmsQueries/__tests__/recentlyAddedContentQuery.spec.js
```diff
@@ -0,0 +1,94 @@
+const { getPagination } = require('../../../utils/jsonApi');
+const { RecentlyAddedContentQuery } = require('../recentlyAddedContentQuery');
+
+describe('Recently Added Content page query', () => {
+  let ESTABLISHMENTNAME;
+  let timeStamp;
```
jonhindle, line 6: These are never reassigned and they are simply types, so should just be a constant:
```
const ESTABLISHMENTNAME = 'Wayland';
const TIMESTAMP = 1651737416;
```

server/repositories/cmsQueries/__tests__/recentlyAddedContentQuery.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+      links = {
+        next: 'URL',
+      };
+    });
+
+    it('should return null when an empty array is provided', () => {
+      response = [];
+
+      const result = query.transform(response, links);
```
jonhindle, line 66: I'd probably just pass the Array.
```
expect(query.transform([], links)).toBeNull();
```

server/repositories/cmsQueries/__tests__/recentlyAddedContentQuery.spec.js
```diff
@@ -0,0 +1,94 @@
...
+              {
+                tile_small: 'AWS_URL',
+              },
+            ],
+            resourceIdObjMeta: {
+              alt: 'IMAGE_ALT_TEXT',
+            },
+          },
+        },
+      ];
```
jonhindle, line 56: Only used once so probably best to declare in that test rather than in the beforeEach?

server/routes/recentlyAdded.js
```diff
@@ -0,0 +1,58 @@
+const express = require('express');
+
+const createRecentlyAddedContentRouter = ({ cmsService }) => {
+  const router = express.Router();
+
+  const body = req => {
```
jonhindle, line 6: Seems an odd name does `body`; its always assigned to related Content, so maybe `getRelatedContent`?

server/utils/__tests__/date.spec.js
```diff
@@ -116,3 +117,33 @@ describe('getDateSelectionFrom', () => {
     ]);
   });
 });
+
+describe('getOffsetUnixTime', () => {
+  let NOW;
+  let NOW_MILLISECONDS;
```
jonhindle, line 123: Sorry my fault, this could just be a const.
```
const NOW_MILLISECONDS = 1577836800;
```

server/utils/__tests__/date.spec.js
```diff
@@ -116,3 +117,33 @@ describe('getDateSelectionFrom', () => {
...
+  it('should handle a real offset value ', () => {
+    expect(getOffsetUnixTime(27, NOW)).toBe(1575504000);
+  });
+
+  it('should handle a negative offset value ', () => {
+    expect(getOffsetUnixTime(-27, NOW)).toBe(1580169600);
+  });
+
+  it('should default to the current date ', () => {
+    jest.useFakeTimers().setSystemTime(new Date('2020-01-01'));
```
jonhindle, line 143: Use the constant.
```
jest.useFakeTimers().setSystemTime(NOW);
```

server/repositories/cmsQueries/__tests__/recentlyAddedContentQuery.spec.js
BarryGeeMOJ: Agreed. However, regarding the format of the expect, assigning the result to the 'result' const makes the expect test more readable. 
`expect(result).toBeNull()`

server/routes/recentlyAdded.js
BarryGeeMOJ: I have updated the 'body' to be named, 'getMostRecentContent' as this is provides better context of what's actually happening. In addition, I have updated 'relatedContent' to be named 'mostRecentContent' for the same reason this is then assigned to the 'relatedContent' field in the data JSON that's passed into the view. 

In the context of this route 'relatedContent' doesn't make sense as this content isn't related to anything. However, the collections view that is used by both this route and again in tags.js which refers to 'relatedContent'. Ideally we would rename 'relatedContent' in both tags.js and the view too, to something more generic, like 'content' to avoid confusion in routes that use the view but aren't displaying 'related content'.

server/utils/__tests__/date.spec.js
BarryGeeMOJ: Agreed, also made NOW a constant as this is the same value in each test.

server/utils/__tests__/date.spec.js
BarryGeeMOJ: Done.

server/routes/recentlyAdded.js
jonhindle: Agreed, do you want to raise a _tidy up_ ticket for it?

server/utils/__tests__/date.spec.js
jonhindle: So the date could be mutated so I'd leave that as a _let_ because we're risking test bleed. I don't think you can _freeze_ a date object? Well not in a conventional way.

jonhindle (approved):
lgtm

PR #500, opened 2022-06-21
1048 feature/allow a user to see when a tile is new

BarryGeeMOJ:
### Context

https://trello.com/c/mTNr3CTH/1048-allow-a-user-to-see-when-a-tile-is-new

n/a

### Intent

- Add the 'new' tag to tiles with content that is 2 days old or newer
- Update and add new supporting tests

Screenshot 2022-06-21 at 13 03 02

### Considerations

n/a

n/a

### Checklist

- [ ] Tested in Development


assets/sass/components/_content-tile-small.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-hub-contentTileSmall {
+  display: flex;
+  flex-wrap: nowrap;
+  justify-content: space-between;
+  align-items: center;
+  margin: 30;
```
jonhindle, line 6: can we use a gov uk value for consistency?

assets/sass/components/_content-tile-small.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-hub-contentTileSmall {
```
jonhindle, line 1: The rest of the tile CSS is in `_home-content.scss` rightly or wrongly. (A victim of evolution I think)
Should we try and keep this CSS together?

BarryGeeMOJ: i agree the tile css should be kept together. Whether that should be in the home-content scss is another matter. As the tiles are used in other areas i think a tile specific scss is a better approach. Happen to discuss this further.

assets/sass/components/_content-tile-small.scss
jonhindle: This class `govuk-!-margin-6` should cover it

BarryGeeMOJ: turns out the margin style isn't actually needed

assets/sass/components/_content-tile-small.scss
BarryGeeMOJ: relocated class into home-content. I think this needs refactoring to break tile styles out into its own file.

jonhindle (approved):
lgtm

PR #501, opened 2022-06-22
feature/998-homepage-three-panel-header

jonhindle:
### Context

https://trello.com/c/K2Hkxiw8/998-create-the-updates-section?filter=label:Front+End

n/a

### Intent

The update block with three columns is added to the new-home.html

not really

### Considerations

n/a

n/a

### Checklist

- [ ] Tested in Development

### DIFF
--- a/assets/sass/components/_home-content.scss
+++ b/assets/sass/components/_home-content.scss
@@ -262,3 +262,14 @@
   justify-content: space-between;
   align-items: center;
 }
+
+.hub-content-panel {
+  box-sizing: border-box;
+  background-color: $real-dark-grey;
+}
+
+.hub-horizontal-rule-title {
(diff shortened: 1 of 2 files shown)


server/routes/homepage.js
```diff
@@ -64,6 +64,7 @@ const createHomepageRouter = ({ cmsService, offenderService }) => {
         },
         hideSignInLink: true,
         title: 'Home',
+        updates: recentlyAddedHomepageContent.data,
```
BarryGeeMOJ, line 67: This probably isn't needed until we know the actual query and we should deconstruct the data for each of the update panels when we have that data and pass those into the view?

jonhindle: Sorry, left over from the dummy mockups 👍

BarryGeeMOJ (approved):

PR #505, opened 2022-06-30
1077 feature/populate the new homepage updates guides section

BarryGeeMOJ:
### Context

https://trello.com/c/kIt9J5J7/1077-populate-the-new-homepage-updates-guides-section

n/a

### Intent

- Added a new component to render key info data within the new homepage updates 
- Added a cropTextWithEllipsis utility function to crop strings to a specified length and to add an ellipse to the end of the cropped string when required
- supporting unit tests for feature added

Screenshot 2022-06-30 at 10 21 12

### Considerations

No.

No.

### Checklist

- [ ] Tested in Development


server/routes/__tests__/homepage.spec.js
```diff
@@ -485,13 +507,53 @@ describe('GET /', () => {
...
+        }));
+
+    it('Should render a html h3 tag with the expected text in the homepage key info content tile', () =>
+      request(app)
+        .get('/new-homepage')
+        .expect(200)
+        .then(response => {
+          const $ = cheerio.load(response.text);
+          expect($('#keyInfo a div div h3:last').text()).toBe(
+            'A title that is long enough to be cropped with an ellipse added to the end',
```
jonhindle, line 546: Personal preference, but this would be nicer as a const but meh:
```
const LAST_TILE_TITLE= 'A title that is long enough to be cropped with an ellipse added to the end'


// ...

expect($('#keyInfo a div div h3:last').text()).toBe(LAST_TILE_TITLE)
```

server/utils/__tests__/jsonApi.spec.js
```diff
@@ -557,3 +558,76 @@ describe('isNew', () => {
...
+      title: 'A title that is long enough...',
+      summary: 'A summary',
+      contentUrl: '/content/999999',
+      displayUrl: undefined,
+      image: {
+        url: 'url path',
+        alt: 'Alt text',
+      },
+      isNew: false,
+    };
```
jonhindle, line 596: Fair bit of duplication. I'd have been tempted to cheat a bit:
```
outputItem = {
  ...item,
  title: 'A title that is long enough...',
}
```

server/utils/__tests__/jsonApi.spec.js
```diff
@@ -557,3 +558,76 @@ describe('isNew', () => {
...
+    expect(title).toStrictEqual(item.title);
+  });
+
+  it("should return an item object with only the 'title' value updated", () => {
+    expect(cropTextWithEllipsis(item, 30)).toStrictEqual(outputItem);
+  });
+
+  it('should return the item provided without changes when a maxNumberOfChars value is not provided', () => {
+    expect(cropTextWithEllipsis(item)).toStrictEqual(item);
+  });
```
jonhindle, line 620: I'm not sure of the value of this. I think I'd want it to error if no limit is provided because its being used for the wrong reasons?

server/utils/jsonApi.js
```diff
@@ -15,6 +15,26 @@ const getLargeImage = data => getImage(data, 'tile_large');
...
 const isNew = fromDate => differenceInDays(new Date(), new Date(fromDate)) <= 2;
 
+const cropTextWithEllipsis = (item, maxNumberOfChars) => {
+  if (!item || JSON.stringify(item) === '{}') {
+    throw new Error('An item object with the expected structure is required');
+  }
+
+  if (!maxNumberOfChars) {
+    return item;
+  }
```
jonhindle, line 25: This masks an error in the code. This function should *never* be called without a number of chars. Unless there is a need to set a default that is? 🤔

server/views/components/key-info-items/template.njk
```diff
@@ -0,0 +1,35 @@
+{% for item in params.data %}
+  <a data-featured-tile-id="{{item.id}}"
```
jonhindle, line 2: The outer CSS class for this component sits outside the component.
The `<div class=".key-info-item">` should really be in this template, if we want the CSS to be linked with this template.

server/views/pages/home-new.html
```diff
@@ -41,10 +42,11 @@ <h2 class="govuk-heading-l govuk-hub-white-heading govuk-!-margin-top-5 govuk-!-
...
       </div>
     </div>
-    <div class="govuk-grid-column-one-quarter govuk-!-padding-0">
-      <!-- 300px column, updates small guide links -->
-      Guide links
-    </div>
+    {% if keyInfo.data | length %}
+      <div class="govuk-grid-column-one-quarter" id="keyInfo">
+        {{ keyInfoItems(keyInfo) }}
+      </div>
```
jonhindle, line 48: See comment above for the component. The `<div` tag should be in the component because otherwise the CSS rules are dependent on both template *and* component. I think they really ought to apply to the component in isolation.

server/utils/jsonApi.js
```diff
@@ -15,6 +15,26 @@ const getLargeImage = data => getImage(data, 'tile_large');
...
+
+  const title =
+    item.title.length > maxNumberOfChars
+      ? `${item.title
+          .substr(0, maxNumberOfChars)
+          .match(/^.*[\w\d]+(?=[^\w\d]+[\w\d]*)/g)[0]
+          .substr(0, maxNumberOfChars - 1)}...`
+      : item.title;
```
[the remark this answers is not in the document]
BarryGeeMOJ: This was based on the code you provided and my understanding is that both would produce the same result. Let's talk about this though.

Assuming in this case num = 30

The code you provided was:

```
title.substr(0, num-2) // (0, 28)
  .match(/^.*[\w\d]+(?=[^\w\d]+[\w\d]*)/g)[0]
  .substr(0, num-3); // (0, 27)
```

```
${item.title
          .substr(0, maxNumberOfChars) // (0, 30)
          .match(/^.*[\w\d]+(?=[^\w\d]+[\w\d]*)/g)[0]
          .substr(0, maxNumberOfChars - 1)}...` // (0, 29)
      : item.title;
```

We can make this exactly the same as the first option by changing maxNumberOfChars to 28, 

```
${item.title
          .substr(0, maxNumberOfChars) // (0, 28)
          .match(/^.*[\w\d]+(?=[^\w\d]+[\w\d]*)/g)[0]
          .substr(0, maxNumberOfChars - 1)}...` // (0, 27)
      : item.title;
```

Providing a variable and only ever reducing that and never using its original value doesn't make sense. 
The current 30/33 including ellipse doesn't need to be an exact thing. There is room for manoeuvre in that value, but 30 characters / 33 including the ellipse appears to fit and work well in the current design.

I think this piece of code is fit for purpose, but I may be misunderstanding the point you're making.

server/utils/jsonApi.js
BarryGeeMOJ: I agree the function 'should' never be called without maxNumberOfChars, this was coded in this way as a simple and none breaking method to handle that eventuality. The way I would change this would be to throw an error instead but I think its valid code to catch this in some form otherwise the function would just blow up.

server/utils/__tests__/jsonApi.spec.js
BarryGeeMOJ: See comment below.

server/utils/__tests__/jsonApi.spec.js
BarryGeeMOJ: good catch.

server/routes/__tests__/homepage.spec.js
BarryGeeMOJ: I think in this case there is no real value from assigning the string to a variable. If this string was used in a number of tests then I would be setting it in a beforeEach and referencing the variable as you suggest. This string is used in a couple of the mock hubContent objects so there may be some value in doing this but prior to the hubContent being defined as apposed to within this test itself.

server/views/components/key-info-items/template.njk
BarryGeeMOJ: this is completely valid and I agree.

server/views/pages/home-new.html
BarryGeeMOJ: again, valid and I agree.

BarryGeeMOJ: updated.

server/views/components/key-info-items/template.njk
BarryGeeMOJ: as below

server/utils/__tests__/jsonApi.spec.js
BarryGeeMOJ: test updated to match updated function as below.

server/utils/jsonApi.js
BarryGeeMOJ: updated to use default value.

server/utils/__tests__/jsonApi.spec.js
BarryGeeMOJ: updated

server/utils/jsonApi.js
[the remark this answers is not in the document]
jonhindle: We do use the max value, the whole piece of code is as follows:
```
const title =
  item.title.length > maxNumberOfChars
    ? `${item.title
      .substr(0, maxNumberOfChars-2)
      .match(/^.*[\w\d]+(?=[^\w\d]+[\w\d]*)/g)[0]
      .substr(0, maxNumberOfChars - 3)}...`
    : item.title;
```
You need to take into account that the ellipses themselves represent *three* characters. 
I mean also this is edge case stuff, and we're not using a monospaced font, so the max chars is already a best guess scenario.

server/views/components/key-info-items/template.njk
```diff
@@ -0,0 +1,37 @@
+<div class="govuk-grid-column-one-quarter" id="keyInfo">
```
jonhindle, line 1: I might have got this wrong 🤔 
Because the CSS uses the class `key-info-item` not the tags. My bad.

jonhindle (approved):
lgtm

PR #544, opened 2022-09-27
feature/1390-connect-up-redis

jonhindle:
### Context

https://trello.com/c/dupoKPmL/1390-implement-redis-for-caching-drupal-requests-in-the-frontend

Most calls are not cached or cached inmemory.

### Intent

Redis setup and content cache rejig.

n/a

### Considerations

n/a

Redis cache needs enabling/creating in the target environment.

### Checklist

- [ ] Tested in Development


server/repositories/cmsApi.js
```diff
@@ -71,6 +76,18 @@ class CmsApi {
     });
   }
 
+  async getCache(query) {
+    const key = query.getKey?.() || null;
+    const expiry = query.getExpiry?.() || 8640;
+    let data = key ? await this.#cache.get(key) : null;
+    if (!data) {
+      data = (await this.get(query)) || [];
+      if (key && expiry && data) await this.#cache.set(key, data, expiry);
```
andrewrlee, line 85: key and expiry won't ever be null will they?

jonhindle: Key theoretically might be - but thats an error, so I probably could/should throw an error in that scenario instead 🤔

andrewrlee: but data would then be null?

server/utils/caching/redisClient.js
```diff
@@ -45,6 +43,7 @@ class RedisCachingStrategy {
   }
 
   async get(key) {
+    logger.info(`Getting cache for '${key}'.`);
```
andrewrlee, line 46: this is going generate a lot of logs, it shouldn't be necessary as you should be able to see the redis calls in app insights as `dependencies`

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+      - hmpps
+    container_name: hub-frontend-redis
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
+    ports:
+      - '6379:6379'
+
+networks:
+  hmpps:
+
```
leonk, line 16: We already have a docker-compose setup in the parent project. https://github.com/ministryofjustice/prisoner-content-hub/blob/main/docker-compose.yml (with Redis enabled for Drupal there).
I wonder if we should move this stuff into there?  That way there is only one docker compose file to maintain.  (I know that you guys don't use this yet, but maybe good if you started to? Also now that we have automated db backups for Drupal, we could probably incorporate that into the local build, as that was always the blocker before).

server/repositories/cmsQueries/categoryContentQuery.js
```diff
@@ -26,6 +27,14 @@ class CategoryContentQuery {
     this.query = `${queryWithoutOffset}&${getPagination(page, limit)}`;
   }
 
+  getKey() {
+    getCmsCacheKey('categoryContent', this.establishmentName);
+  }
+
+  getExpiry() {
+    return 3600;
```
leonk, line 35: This is content on a category page?  I think caching for one hour will cause confusion (i.e. I added a piece of content, where is it?  Or I had to remove something, but it's still there).

The one to cache would be the sub_terms request.  As this just displays series and subcategories, and so won't need to be updated as regularly.  Also DCM's don't add new series/categories, only studio team do, and they can have a better understanding about the caching.

server/repositories/cmsApi.js
jonhindle: I threw an error for missing keys.

server/repositories/cmsQueries/categoryContentQuery.js
jonhindle: I think we go through each call and agree what caches and for how long.

BarryGeeMOJ (approved):
lgtm

PR #562, opened 2022-11-08
Adding docker compose environment setup.

leonk:
### Context

https://trello.com/c/66Bk2qCz/1716-spike-improve-local-environment-setup

### Intent

Improve the local environment setup by using docker compose.
There's a PR open for the backend repo https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/476 this will bring in the docker-compose setup out of the parent repo (http://github.com/ministryofjustice/prisoner-content-hub).  Instead we will have two separate docker-compose.yml files (one in the backend and one in the frontend).

They will each be on the same docker network `prisoner-content-hub`, they will therefore be able to communicate with each other.

**Note** this PR is a draft as we still need to determine what to do about the existing [.env-template](https://github.com/ministryofjustice/prisoner-content-hub-frontend/blob/main/.env-template) file and also update the README with the new setup.

### Considerations
The only issue with re-using the same network (that I can tell) is that if you have both environments running, when you try to stop one of them you get the following error:
```
failed to remove network prisoner_content_hub: Error response from daemon: error while removing network: network prisoner_content_hub id e3a5c58c1d5efd5e865dd0c41b503e8a38370c65af00d63b0400ac90693de587 has active endpoints
```

### Checklist

- [ ] Tested in Development


docker-compose.yml
```diff
@@ -1,16 +1,25 @@
-version: '3.1'
+version: "3.5"
```
leonk, line 1: Apparently to use a named network you need to use version 3.5.  See https://stackoverflow.com/a/48024244

docker-compose.yml
```diff
@@ -1,16 +1,25 @@
...
     image: 'bitnami/redis:5.0'
     networks:
-      - hmpps
+      - prisoner_content_hub
     container_name: hub-frontend-redis
     environment:
       - ALLOW_EMPTY_PASSWORD=yes
     ports:
-      - '6379:6379'
+      - '6378:6378'
```
leonk, line 21: I've had to update the port number here as 6379 is already in use for the backend Redis instance.

prisoner-content-hub-frontend.env
```diff
@@ -0,0 +1,24 @@
+PORT=3000
+
+# The following must match the container names in your docker-compose or local setup
+HUB_API_ENDPOINT=http://drupal:8080
```
leonk, line 4: I think maybe we'd want the default to be the Drupal development environment?  And then a way to optionally override this with a local version of Drupal?  Upto you @BarryGeeMOJ

prisoner-content-hub-frontend.env
```diff
@@ -0,0 +1,24 @@
...
+PRISON_API_BASE_URL='http://foo.bar'
+HMPPS_AUTH_BASE_URL='http://foo.bar'
+HMPPS_AUTH_CLIENT_ID='UNSET'
+HMPPS_AUTH_CLIENT_SECRET='UNSET'
+
+ENABLE_FEATURE_TOGGLES=true
+ENABLE_STACK_TRACES_ON_ERROR_PAGES=true
+ENABLE_MOCK_AUTH=true
+ENABLE_REDIS_CACHE='false'
+
```
leonk, line 24: With this file, I'm not exactly sure what is in use and what isn't.  So it would be worth going through it and making sure we are only setting env variables that we actually need.

The backend will also have a separate `prisoner-content-hub-backend-local.env` file, which will be in git but also in .gitignore (so the file is created with defaults on a git checkout, buy any local amendments do not go back into the repo automatically).  We may want to copy that setup here?

prisoner-content-hub-frontend.env
```diff
@@ -0,0 +1,24 @@
+PORT=3000
+
+# The following must match the container names in your docker-compose or local setup
+HUB_API_ENDPOINT=http://drupal:8080
+ELASTICSEARCH_ENDPOINT=http://elasticsearch:9200
+ELASTICSEARCH_INDEX_NAME='content_for_search'
+DRUPAL_DATABASE_NAME='hubdb'
+
+ENABLE_PERSONALISATION=false
+ENABLE_REDIS_CACHE=false
```
leonk, line 10: Shall we enable Redis caching by default? (Probably yes)

docker-compose.yml
```diff
@@ -1,16 +1,25 @@
...
+version: "3.5"
 services:
-
-  hub-frontend-redis:
+  frontend:
+    image: quay.io/hmpps/prisoner-content-hub-frontend
+    build:
+      context: .
+    env_file:
+      - prisoner-content-hub-frontend.env
```
psoleckimoj, line 8: Can't you just pass envs in when you start this up like redis is doing? It's how we normally do it, i.e. https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/docker-compose.yml We just pass env vars in.

psoleckimoj: I.e. `--env-file ./config/.env.dev` option for starting up docker compose.

leonk: Yeah can just pass them straight in.  The reason for having a separate file, is that on the backend we have separate files (so basically for consistency).  

I think it makes sense for the backend as it would make the `docker-compose.yml` file too large, considering there are 6 services in the [docker-compose.yml](https://github.com/ministryofjustice/prisoner-content-hub-backend/blob/5762a0c0981804fa054290d996620a6eda415c51/docker-compose.yml) file and about 30 env variables just for the backend https://github.com/ministryofjustice/prisoner-content-hub-backend/blob/5762a0c0981804fa054290d996620a6eda415c51/prisoner-content-hub-backend.env

psoleckimoj: 👍 understood, the docs have the env vars in anyways. Which is all fine, it doesn't actually change how it works for local dev if one chooses since you can pick and choose which services to start in the compose 🙂

PR #647, opened 2023-04-21
Refactor `Sentry` configuration to reduce transaction consumption

connormaglynn:
### Context

No

NA

### Intent

This will reduce overall Sentry Transaction usage by approximately by 75%, with no impact on Errors being sent to Sentry.

This is done by ignoring the **POST /analytics/page'** and  **POST /analytics/event'**. These endpoints are a technical workaround to send data to Google Analytics, so presumably, there's little value in the Transaction being sent to Sentry, and **POST /analytics/page'** is sent for every page load within the system, so using a disproportionate amount of transactions.

This PR also reduces the total sampling rate from 25% to 5%.

With both proposed changes, we're estimated to save 5million transactions being sent to Sentry, with minimal impact on value.

Again, if value is lost in these changes - I'd be more than happy for it not to be approved! 😄 

NA

### Considerations

The quota for Sentry usage is MoJ-wide. If the quota is reached for the month, Sentry will stop recording transactions for everything. Therefore, it's beneficial for all product teams to reduce consumption where possible if there is little to no value in the consumption.

No

### Checklist

- [ ] Tested in Development

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -69,9 +69,11 @@ const createApp = services => {
       return transactionName &&
         (transactionName.includes('/health') ||
           transactionName.includes('/public/') ||
-          transactionName.includes('/assets/'))
+          transactionName.includes('/assets/') ||
+          transactionName.includes('/analytics/page') ||
+          transactionName.includes('/analytics/event'))
         ? 0
-        : 0.25;
+        : 0.5;
     },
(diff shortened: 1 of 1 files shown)


psoleckimoj (approved):
LGTM!

PR #667, opened 2023-10-03
LNP-286: 🔧 use opensearch instead of elasticsearch to post feedback.

Eli-TW:
### Context

> Does this issue have a Jira ticket?

https://dsdmoj.atlassian.net/browse/LNP-286

Go to a page on the content hub. Give feedback on that page. That feedback is not stored anywhere.

### Intent

🔥  remove unused variables such as Drupal DB name.
🔧  derive the endpoint to which to save feedback from k8s secret exported from the new opensearch resource

No

### Considerations

No

No

### Checklist

- [x] Tested in Development


server/config.js
```diff
@@ -9,11 +9,10 @@ const elasticsearchEndpoint = getRequiredEnv(
   'ELASTICSEARCH_ENDPOINT',
   'http://localhost:9200',
 );
-const elasticsearchIndexName = getRequiredEnv(
-  'ELASTICSEARCH_INDEX_NAME',
-  'content_index',
+const feedbackEndpoint = getRequiredEnv(
+  'FEEDBACK_ENDPOINT',
```
psoleckimoj, line 13: Are you going to change this for each environment in the Helm values-* files?

Eli-TW: I wasn't intending to.

This variable was the same in all environments before any changes so I have kept it consistent. Each front end environment will be pointing at a different instance of opensearch so we don't need to vary this path to segregate feedback from different environments.

I presume this is a hangover from an idea that may or may not have ever been implemented to use a single elasticsearch instance for feedback from all environments. It could be clearer at this point to change the endpoint to /feedback/_doc in all environments, but that would then require additional changes on the collection mechanism.

psoleckimoj: No worries, it was just with it having `local` in the name that was all.

psoleckimoj (approved):
LGTM!

PR #675, opened 2023-12-04
Fix money routes processSelectedDate

psoleckimoj:
### Context

> Does this issue have a Jira ticket?

No.

### Intent

> What changes are introduced by this PR that correspond to the above ticket?

date function allowed dependency injection so worked for testing, however code that used it did not allow dependency injection.

### Considerations

### Checklist

- [ ] This PR contains **only** changes related to the above ticket
- [ ] Tested in Development

### DIFF
--- a/server/routes/money.js
+++ b/server/routes/money.js
@@ -1,5 +1,11 @@
 const express = require('express');
-const { startOfMonth, parseISO, endOfMonth, isFuture } = require('date-fns');
+const {
+  startOfMonth,
+  parseISO,
+  endOfMonth,
+  isFuture,
+  isValid,
+} = require('date-fns');
 const {
   createTransactionsResponseFrom,
   createDamageObligationsResponseFrom,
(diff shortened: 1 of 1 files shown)


server/routes/money.js
```diff
@@ -15,14 +21,22 @@ function isValidDateSelection(selectedDate, dateSelection) {
 }
 
 function processSelectedDate(selectedDate) {
-  const dateSelection = getDateSelection(new Date(), parseISO(selectedDate));
+  const parsedDate = parseISO(selectedDate);
```
psoleckimoj, line 24: I have no idea why this suddenly started to fail. The failing tests were calculating incorrect dates due to trying to use today's date with previous years causing a weird set of results. The underlying function `getDateSelection` is fine and allows for dependency injection but this cannot happen here due to it being a route that's being tested.

I have pulled the current year out of the selectedDate (if parsable) to make `new Date()` behave as if it's the year from the selectedDate rather than when the tests are run.

PR #724, opened 2024-10-07
Disable capturing of inproc dependencies in app insights

andrewrlee:
Sentry was registering extra events against app insights.


madebyzak (approved):
LGTM!

PR #806, opened 2025-08-19
LNP-1219: Add CSP rules

chriswaugh:
### Context

LNP-1219: Helmet config missing CSP config - highlighted by Github scan

### Intent

Added a loose content security policy to give some basic protection and fix the issue github code scanning is reporting.

### Considerations

We ideally would want to have a more strict CSP and not allow `unsafe-inline` scripts. However this would require more work such as adding randomly generated nonces to inline scripts and styles.

Question is - do we want to do this now and expand the scope beyond just fixing the scan warning, or do we want to add a new item to the backlog for enhancing the CSP beyond what's covered here?

### Checklist

- [x] This PR contains **only** changes related to the above ticket
- [x] Tests have been added/updated to cover the change (N/A - config change so tested manually)
- [x] Tested in Development

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -81,7 +81,7 @@ const createApp = services => {
           defaultSrc: ["'self'"],
           scriptSrc: ["'self'", "'unsafe-inline'", 'www.googletagmanager.com'],
           imgSrc: ["'self'", '*.amazonaws.com'],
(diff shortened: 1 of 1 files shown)


angela-pan-moj (approved):
I think at this point fixing the warning is good enough, but will add a backlog item to look at enhancing the CSP further.

